PR #5, opened 2017-09-22
Check directly if mergeable field is present

ungureanuvladvictor:
Fixes the case when the PR is not mergeable (value being false) and `bulldozer` spins up forever querying for the status.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -392,20 +392,24 @@ func (client *Client) PullRequestForSHA(repo *github.Repository, SHA string) (*g
 		return nil, nil
 	}
 
+	if pullRequest.Mergeable != nil {
+		return pullRequest, nil
+	}
+
 	ticker := time.NewTicker(5 * time.Second)
 	for {
+		select {
+		case <-ticker.C:
+		}
+
 		// polling for merge status (https://developer.github.com/v3/pulls/#get-a-single-pull-request)
 		p, _, err := client.PullRequests.Get(client.Ctx, owner, name, pullRequest.GetNumber())
 		if err != nil {
 			return nil, errors.Wrapf(err, "cannot poll for PR merge status with head %s on %s", SHA, repo.GetFullName())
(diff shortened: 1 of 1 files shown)


github/github.go
```diff
@@ -392,21 +392,25 @@ func (client *Client) PullRequestForSHA(repo *github.Repository, SHA string) (*g
...
+		case <-ticker.C:
+		}
+
 		// polling for merge status (https://developer.github.com/v3/pulls/#get-a-single-pull-request)
 		p, _, err := client.PullRequests.Get(client.Ctx, owner, name, pullRequest.GetNumber())
 		if err != nil {
 			return nil, errors.Wrapf(err, "cannot poll for PR merge status with head %s on %s", SHA, repo.GetFullName())
 		}
-		if p.GetMergeable() {
+		if p.Mergeable != nil {
```
bluekeyes: Now return the pull request whether it is mergeable or not, whereas before,we only returned it if was mergeable. Do we check the mergeability of the PR again in one of the callers of this function before taking action? If not, do we even need to wait for `p.Mergeable` to be non-nil?

ungureanuvladvictor: It is checked at https://github.com/palantir/bulldozer/blob/master/server/endpoints/hook.go#L202.

github/github.go
```diff
@@ -392,21 +392,25 @@ func (client *Client) PullRequestForSHA(repo *github.Repository, SHA string) (*g
...
 	}
 
+	if pullRequest.Mergeable != nil {
+		return pullRequest, nil
+	}
+
 	ticker := time.NewTicker(5 * time.Second)
 	for {
+		select {
+		case <-ticker.C:
```
bluekeyes: Oh, forgot to mention it before - this `select` isn't necessary, as there is only one condition. You can just do `<-ticker.C` to block until a tick arrives instead.

PR #16, opened 2017-09-30
Moving to use golang/dep

ungureanuvladvictor:
Thinking about moving to use `golang/dep` as dependency manager for `bulldozer`.


ajlake:
I'm not sure where the +2.3 million LOC comes from. Swapping vendor tools should be roughly net zero so where is the discrepancy coming from? How do the built binary sizes compare?

bluekeyes:
Did you run `dep prune` before committing the vendor directory?

ungureanuvladvictor:
Looks like that brought down the LOC count.

Issue #27, opened 2017-12-18
Bulldozer should only require github reviews if github reviews are required


ryanmcnamara:
Currently if any reviewer has not approved the pr, bulldozer will not merge the pr. 

There is a preference specifically for this behavior (Require pull request reviews before merging) and when it isn't enabled, bulldozer shouldn't enforce this

PR #28, opened 2017-12-18
Modify github approval verification logic

ryanmcnamara:
If reviews are not required for a repo, then only consider review status
as failed if someone has explicitly disapproved the repo

Fixes https://github.com/palantir/bulldozer/issues/27


ryanmcnamara:
Not sure how to test this

PR #36, opened 2018-02-24
Create bulldozer labels at repo enable time

### LINKED ISSUE #25: Create labels if they don't exist
Bulldozer is configured via labeled PRs, but a human has to create all those labels by hand. If I enable bulldozer on a repo which does not have the labels, they should be created for me.

ungureanuvladvictor:
- Removes global variables `GithubOauthConfig` and `Instance`
- Creates `update me`, `merge when ready` and `do not merge` labels if they do not exist
- Add `deliveryID` on the logger only if it exists in the request to the server

Fixes #25

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/36)

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -40,8 +40,6 @@ var serverCmd = &cobra.Command{
 			log.Fatalf("Failed to parse config file: %+v", err)
 		}
 
-		config.Instance = startup
-
 		if startup.LogLevel() == log.DebugLevel {
 			cmd.DebugFlags()
 			viper.Debug()
@@ -53,7 +51,7 @@ var serverCmd = &cobra.Command{
 			log.Fatalf("Cannot init db: %+v", err)
 		}
 
-		auth.InitGithubAuth(startup.Github)
+		auth.GitHubOAuth(startup.Github)
 
 		srv := server.New(db, startup)
 		if err := srv.SetupSessionStore(); err != nil {
--- a/github/github.go
+++ b/github/github.go
@@ -69,7 +69,7 @@ type BulldozerFile struct {
 	Mode             string `yaml:"mode" validate:"nonzero"`
 }
 
-func FromToken(c echo.Context, token string) *Client {
+func FromToken(c echo.Context, ghAPIURL, token string) *Client {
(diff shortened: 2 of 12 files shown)


cmd/server.go
```diff
@@ -53,7 +51,7 @@ var serverCmd = &cobra.Command{
 			log.Fatalf("Cannot init db: %+v", err)
 		}
 
-		auth.InitGithubAuth(startup.Github)
+		auth.GitHubOAuth(startup.Github)
```
bluekeyes: You can delete this line since `GitHubOAuth()` has no side-effects and the result is unused.

github/github.go
```diff
@@ -676,3 +676,33 @@ func (client *Client) AllRepositories(user *github.User) ([]*github.Repository,
 
 	return ownedRepos, nil
 }
+
+func (client *Client) CreateLabels(repo *github.Repository) error {
+	updateMeLabel := config.UpdateMeLabels[0]
+	doNotMergeLabel := config.DoNotMergeLabels[0]
+	mergeWhenReadyLabel := config.MergeWhenReadyLabels[0]
```
bluekeyes: Right now, it looks like if there are multiple labels that are accepted for each condition, only the first one is checked/created here. If a repository already has the secondary labels, then this will lead to "duplicate" labels (different labels that have the same behavior).

I think we need to either:

1.  Account for this by checking for all possible labels, and only creating the first one if none exist
2. Change the configuration so that there can only be one label for each condition

server/endpoints/repositories.go
```diff
@@ -198,6 +198,10 @@ func RepositoryEnable(db *sqlx.DB, webHookURL string, webHookSecret string) echo
 			return errors.Wrapf(err, "cannot add %s/%s to the database", owner, name)
 		}
 
+		if err := client.CreateLabels(repo); err != nil {
+			return errors.Wrapf(err, "cannot create Bulldozer issues on repo %s", repo.GetFullName())
```
bluekeyes: `Bulldozer labels on repo`

PR #45, opened 2018-07-16
Add option to ignore squashed commit messages

robert3005:
This is partially a nice to have and partially workaround of github pr comment body generation. 

For short prs it's sometimes enough to just have a title and no description. Then you don't need to add anything to the pr comment to get a clean commit message on merge.

At the same time if you're creating prs via api and not through ui the pr template is not applied and you have to manually include it. It's better to be explicit in those cases instead of letting bulldozer add all squashed commit messages to the commit body.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/45)


robert3005:
This isn't the best behaviour - need to handle the case if `==SQUASH_MSG==` is specified since right now we will flat out ignore it

PR #54, opened 2018-11-02
Add dist configuration for Godel

bluekeyes:
This is required for Bintray uploads to work. Since there's only a
single binary, os-arch-bin might be more appropriate, but this matches
policy-bot for now.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/54)

### DIFF
--- a/godel/config/dist-plugin.yml
+++ b/godel/config/dist-plugin.yml
@@ -4,8 +4,6 @@ product-defaults:
 
 products:
   bulldozer:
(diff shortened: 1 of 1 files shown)


bluekeyes:
Because this was missing, the 1.0.0 publish never worked. I manually published 1.0.0 from a Circle build while debugging this.

asvoboda (approved):
:/

PR #56, opened 2018-11-05
Remove close PR calls, update required permissions

### LINKED ISSUE #52: Merged pull requests are not closed when using 1.0
On some repositories, after Bulldozer merges a PR, GitHub Enterprise 2.14.6 still sees the state of the PR as open: it shows up in open PR lists and has a button to close it. The merge is complete and the commits exist in the target branch, so it is only the PR state that is incorrect.

We have an open support issue with GitHub for this, since it seems like an internal race condition or an issue with app merges, rather than a bug with Bulldozer. The 0.4.x OAuth app never had this problem despite using very similar merge code.

If we can't figure out what's wrong, we'll probably need to add a workaround to close PRs via the API after they are merged.

bluekeyes:
GitHub does close PRs, but the app has to have the correct permissions.
While write on the repo contents is needed for merging, write on PRs
seems to be needed for closing them.

This reverts commit 25d3ff129b605185c51de03b7b3f4f8c7d65711f.

This _really_ fixes #52.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/56)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ bulldozer requires the following permissions as a GitHub app:
 * Repository Contents - read & write
 * Issues - read-only
 * Repository metadata - read-only
-* Pull requests - read-only
+* Pull requests - read & write
 * Commit status - read-only
 
 It should be subscribed to the following events:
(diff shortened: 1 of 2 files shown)


asvoboda (approved):
🥇

PR #60, opened 2018-11-13
add back following PR events

jmcampanini:
adds back the handler for `pull_requests`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/60)


asvoboda:
I suspect you'll also need to register the handler: https://github.com/palantir/bulldozer/blob/develop/server/server.go#L73

bluekeyes (approved):

PR #61, opened 2018-11-13
Disable updates if there is no whitelist or blacklist

jmcampanini:
bug: if there was no whitelist and no blacklist, then all branches were updated. the update behavior should be opt-in.

this should:

* have simple logic in `ShouldUpdatePR`
* tests to prove all cases work correctly

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/61)

### DIFF
--- a/bulldozer/update_test.go
+++ b/bulldozer/update_test.go
@@ -16,6 +16,7 @@ package bulldozer
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -27,11 +28,11 @@ import (
 func TestShouldUpdatePR(t *testing.T) {
 	ctx := context.Background()
 	testMatrix := []struct {
-		blacklistable   bool
-		blacklisted     bool
-		whitelistable   bool
-		whitelisted     bool
-		expectingUpdate bool
+		blacklistEnabled bool
+		blacklisted      bool
+		whitelistEnabled bool
+		whitelisted      bool
+		expectingUpdate  bool
 	}{
 		{false, false, false, false, false},
 		{false, false, false, true, false},
@@ -52,13 +53,15 @@ func TestShouldUpdatePR(t *testing.T) {
 	}
 
 	for ndx, testCase := range testMatrix {
-		pullCtx, updateConfig := generateUpdatableTestCases(testCase.blacklistable, testCase.blacklisted, testCase.whitelistable, testCase.whitelisted)
+		pullCtx, updateConfig := generateUpdateTestCase(testCase.blacklistEnabled, testCase.blacklisted, testCase.whitelistEnabled, testCase.whitelisted)
 		updating, err := ShouldUpdatePR(ctx, pullCtx, updateConfig)
-		require.Equal(t, testCase.expectingUpdate, updating, "case %d", ndx)
 		require.NoError(t, err)
+		msg := fmt.Sprintf("case %d - blacklistEnabled=%t blacklisted=%t whitelistEnabled=%t whitelisted=%t -> doUpdate=%t",
+			ndx, testCase.blacklistEnabled, testCase.blacklisted, testCase.whitelistEnabled, testCase.whitelisted, testCase.expectingUpdate)
+		require.Equal(t, testCase.expectingUpdate, updating, msg)
 	}
 }
-func generateUpdatableTestCases(blacklistable bool, blacklisted bool, whitelistable bool, whitelisted bool) (pull.Context, UpdateConfig) {
+func generateUpdateTestCase(blacklistable bool, blacklisted bool, whitelistable bool, whitelisted bool) (pull.Context, UpdateConfig) {
 	updateConfig := UpdateConfig{}
 	pullCtx := pulltest.MockPullContext{}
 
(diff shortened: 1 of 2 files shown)


bluekeyes:
Can you explain the problem this is fixing?

jmcampanini:
updated PR body with bug description.

kept the test and tried to change as little logic as possible, and ended up with the following:

```go
func ShouldUpdatePR(ctx context.Context, pullCtx pull.Context, updateConfig UpdateConfig) (bool, error) {
	logger := zerolog.Ctx(ctx)

	if updateConfig.Blacklist.Enabled() {
		blacklisted, reason, err := IsPRBlacklisted(ctx, pullCtx, updateConfig.Blacklist)
		if err != nil {
			return false, errors.Wrap(err, "failed to determine if pull request is blacklisted")
		}

		if blacklisted {
			logger.Debug().Msgf("%s is deemed not updateable because blacklisting is enabled and %s", pullCtx.Locator(), reason)
			return false, nil
		}

		if !updateConfig.Whitelist.Enabled() { // added
			return true, nil
		}
	}

	if updateConfig.Whitelist.Enabled() {
		whitelisted, reason, err := IsPRWhitelisted(ctx, pullCtx, updateConfig.Whitelist)
		if err != nil {
			return false, errors.Wrap(err, "failed to determine if pull request is whitelisted")
		}

		if !whitelisted {
			logger.Debug().Msgf("%s is deemed not updateable because whitelisting is enabled and no whitelist signal detected", pullCtx.Locator())
			return false, nil
		}

		logger.Debug().Msgf("%s is whitelisted because whitelisting is enabled and %s", pullCtx.Locator(), reason)
		return true, nil // added
	}

	return false, nil // changed
}
```

bulldozer/update_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/palantir/bulldozer/pull"
+	"github.com/palantir/bulldozer/pull/pulltest"
+)
+
+func TestShouldUpdatePR(t *testing.T) {
+	ctx := context.Background()
+	testMatrix := []struct {
+		blacklistable   bool
```
bluekeyes: I think renaming this to `blacklistEnabled` will make more sense (same thing for `whitelistable`).

bulldozer/update_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	for _, testCase := range testMatrix {
+		pullCtx, updateConfig := generateUpdatableTestCases(testCase.blacklistable, testCase.blacklisted, testCase.whitelistable, testCase.whitelisted)
+		updating, err := ShouldUpdatePR(ctx, pullCtx, updateConfig)
+		require.Equal(t, testCase.expectingUpdate, updating)
+		require.NoError(t, err)
+	}
+}
+func generateUpdatableTestCases(blacklistable bool, blacklisted bool, whitelistable bool, whitelisted bool) (pull.Context, UpdateConfig) {
```
bluekeyes: `generateUpdateTestCase`

bluekeyes:
We should make sure that the logic for `ShouldMergePR` doesn't need updating as well. I'd like to make sure the two configurations behave the same if possible.

As for simplifying the code, what about updating a value in place instead of returning immediately?

```go
// take no action by default
update := false

if whitelistEnabled {
  // can merge if it was on the whitelist
  update = whitelisted
}

if blacklistEnabled {
  // or if it was not on the blacklist and there's no whitelist
  update = !blacklisted && !whitelistEnabled
}

return update
```

Or checking that at least one of the lists is enabled upfront? Assuming that the existing code is correct when one or both lists are set.

```go
if !whitelistEnabled && !blacklistEnabled {
  return false
}
// existing code with no changes
```

bulldozer/update_test.go
```diff
@@ -0,0 +1,82 @@
...
+		{true, true, false, true, false},
+		{true, true, true, false, false},
+		{true, true, true, true, false},
+	}
+
+	for _, testCase := range testMatrix {
+		pullCtx, updateConfig := generateUpdatableTestCases(testCase.blacklistable, testCase.blacklisted, testCase.whitelistable, testCase.whitelisted)
+		updating, err := ShouldUpdatePR(ctx, pullCtx, updateConfig)
+		require.Equal(t, testCase.expectingUpdate, updating)
+		require.NoError(t, err)
```
bluekeyes, line 58: I'd check for the error first, before asserting the main condition

bulldozer/update_test.go
```diff
@@ -0,0 +1,82 @@
...
+		{true, true, false, false, false},
+		{true, true, false, true, false},
+		{true, true, true, false, false},
+		{true, true, true, true, false},
+	}
+
+	for _, testCase := range testMatrix {
+		pullCtx, updateConfig := generateUpdatableTestCases(testCase.blacklistable, testCase.blacklisted, testCase.whitelistable, testCase.whitelisted)
+		updating, err := ShouldUpdatePR(ctx, pullCtx, updateConfig)
+		require.Equal(t, testCase.expectingUpdate, updating)
```
bluekeyes: It might be nice if this included a message so we knew which of the combinations failed.

jmcampanini:
solving the explicit case passes all the tests. gonna leave that

bluekeyes (approved):

PR #65, opened 2018-11-14
post comment for user-fixable errors

### LINKED ISSUE #63: comment when encountering an error that is user-resolvable
there are times when bulldozer encounters an error that is user-resolvable, specifically when a merge fails due to permission settings.

we currently just log these, example:

```
Merge rejected due to unsatisfied condition \"You're not authorized to push to this branch. Visit https://help.github.com/enterprise/2.15/user/articles/about-protected-branches/ for more information.\
```

in this particular case, we should be able to add a comment to the PR with feedback on what happened and possible resolution steps

jmcampanini:
fixes #63 

### Current Behavior
There are times where bulldozer will fail to merge a PR due to user error, specifically:

* when the merge strategy configured for bulldozer is not allowed in the repo
* when the bulldozer bot is not allowed to merge to the target branch, due to branch protections

There is no signal to the user that they can fix this issue, so they are left thinking bulldozer is broken.

### Desired Behavior
Bulldozer should post a comment in the 2 above cases, alerting the user to:

* the error returned by GitHub, since it may contain debug information, URLs, etc.
* potential solution to the problem

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/65)

### DIFF
--- a/bulldozer/merge.go
+++ b/bulldozer/merge.go
@@ -107,6 +107,8 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
 					continue
 				}
 
+				postCommentIfUserActionable(ctx, client, pullCtx.Owner(), pullCtx.Repo(), pullCtx.Number(), *gerr)
+
 				switch gerr.Response.StatusCode {
 				case http.StatusMethodNotAllowed:
 					logger.Info().Msgf("Merge rejected due to unsatisfied condition %q", gerr.Message)
@@ -194,4 +196,31 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
 	}
 
 	return repositoryCommits, nil
+}
+
+func postCommentIfUserActionable(ctx context.Context, client *github.Client, owner, repo string, prNum int, githubError github.ErrorResponse) {
+	commentBody := "Error occurred when attempting merge: " + githubError.Message
+
+	switch {
+	case strings.HasPrefix(githubError.Message, "You're not authorized to push to this branch."):
+		commentBody = commentBody + "\n\nThe bulldozer bot account may not be allowed to merge to the branch."
+
+	case githubError.Message == "Merge commits are not allowed on this repository.":
+		commentBody = commentBody + "\n\nThe bulldozer configuration for merge method may not match the repository settings."
+
+	default:
+		// not user-actionable, don't post anything
+		return
+	}
+
+	logger := zerolog.Ctx(ctx)
+	logger.Debug().Msgf("Merge error may be user-fixable, commenting with: %q", commentBody)
+
+	if _, _, err := client.Issues.CreateComment(ctx, owner, repo, prNum, &github.IssueComment{Body: &commentBody}); err != nil {
+		logger.Error().Err(err).Msg("Failed to post user-fixable comment.")
+		return
+	}
+
+	logger.Info().Msgf("Posted comment to PR with potential user action.")
+	return
 }
--- a/bulldozer/merge.go
+++ b/bulldozer/merge.go
@@ -203,7 +203,7 @@ func postCommentIfUserActionable(ctx context.Context, client *github.Client, own
 
 	switch {
 	case strings.HasPrefix(githubError.Message, "You're not authorized to push to this branch."):
-		commentBody = commentBody + "\n\nThe bulldozer bot account may not be allowed to merge to the branch."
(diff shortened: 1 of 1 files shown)


bulldozer/merge.go
```diff
@@ -195,3 +197,30 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
...
 }
+
+func postCommentIfUserActionable(ctx context.Context, client *github.Client, owner, repo string, prNum int, githubError github.ErrorResponse) {
+	commentBody := "Error occurred when attempting merge: " + githubError.Message
+
+	switch {
+	case strings.HasPrefix(githubError.Message, "You're not authorized to push to this branch."):
+		commentBody = commentBody + "\n\nThe bulldozer bot account may not be allowed to merge the branch."
+
+	case githubError.Message == "Merge commits are not allowed on this repository.":
```
bluekeyes, line 208: I think this one could also be something along the lines of `Squash and merge is not allowed` or `Rebase is not allowed`.

bulldozer/merge.go
```diff
@@ -195,3 +197,30 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
 
 	return repositoryCommits, nil
 }
+
+func postCommentIfUserActionable(ctx context.Context, client *github.Client, owner, repo string, prNum int, githubError github.ErrorResponse) {
+	commentBody := "Error occurred when attempting merge: " + githubError.Message
```
bluekeyes, line 202: I would be nice to make this more conversational and nicely formatted, maybe:

```
I wasn't able to merge this pull request! GitHub returned this error:

> {{githubError.Message}}

This can happen when {{reasonThisCanHappen}}.
```

bulldozer/merge.go
```diff
@@ -195,3 +197,30 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
...
+
+	switch {
+	case strings.HasPrefix(githubError.Message, "You're not authorized to push to this branch."):
```
bluekeyes, line 205: Since there are already two cases, it would be nice to provide a bit of abstraction around matching these. Something like this?

```go
type ErrorMatcher struct {
  Pattern regexp.Regexp
  Reason  string
}

type (m *ErrorMatcher) Matches(err github.ErrorResponse) bool {
  // ...
}

var UserErrors = []*ErrorMatcher{
  // ...
}
```

bulldozer/merge.go
```diff
@@ -195,3 +197,30 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
...
+		commentBody = commentBody + "\n\nThe bulldozer bot account may not be allowed to merge the branch."
```
bluekeyes, line 206: As far as I know, this is specifically when branch protection sets restrictions on who can push to a branch and there are no known workarounds. To me, the current wording suggests you can give Bulldozer more permissions to fix this.

bulldozer/merge.go
```diff
@@ -195,3 +197,30 @@ func allCommits(ctx context.Context, pullCtx pull.Context, client *github.Client
...
+
+	default:
+		// not user-actionable, don't post anything
+		return
+	}
+
+	logger := zerolog.Ctx(ctx)
+	logger.Debug().Msgf("Merge error may be user-fixable, commenting with: %q", commentBody)
+
+	if _, _, err := client.Issues.CreateComment(ctx, owner, repo, prNum, &github.IssueComment{Body: &commentBody}); err != nil {
```
bluekeyes, line 219: I think we need to check for any comments by our user or comments with same body to avoid duplicating the warning multiple times on a PR, if other events happen before the problem is resolved (e.g. branch updates, additional non-required status checks reporting, etc.)

jmcampanini: def need to check for previous comments or this can get spammy real quick

bulldozer/merge.go
jmcampanini: yea, the wording should be changed to use a different approach, since bulldozer cant be given more permissions

bulldozer/merge.go
jmcampanini: yea, this sort of logic is best captured in the abstraction you described above.

jmcampanini:
on second thought, im going to add a FAQ section to the README. the complexity and potential endless-loop here makes me think that the first attempt at fixing this should be some documentation. then if it proves to be a real problem, resort to a code change.

PR #69, opened 2018-11-20
Evaluate PR Body independently from comments

### LINKED ISSUE #68: Support for ==MERGE_WHEN_READY== in PR description only
Pre-1.0, in pr_body mode, `==MERGE_WHEN_READY==` was only respected in the PR description, not in comments. For some use cases, this is actually preferred over the change introduced for https://github.com/palantir/bulldozer/issues/31.

For one, we have a bot that comments on PRs that have passed validation that the user can add ==MERGE_WHEN_READY== to the description for it to auto-merge. In 1.0, this comment itself would cause the PR to auto-merge, so we will have to escape that text somehow.

Secondly, updating the PR description is limited to the PR author and those with write access to the repo. We limit write access generally and users submit PRs from forks, so having it be allowed in comments, which are not as locked down, opens up this configuration more broadly than we would like.

Is it possible to configure if comments should be included or not?

asvoboda:
Right now we compare comment substrings on both comments and the PR body. In some cases,
it should be possible to lock down evaluation to only the PR body, instead of comments
_and_ the pr body.

Users who submit PRs do not have write access to the repo, so they cannot add a label.
For comments, anyone with read access to the repo can add a comment on any PR.

However, only someone with write access (or the Author) can edit the PR description.

Fixes #68

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/69)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ merge:
     # "comment_substrings" matches on substrings in comments
     comment_substrings: ["==MERGE_WHEN_READY=="]
 
+    # "pr_body_substrings" matches on substrings in the PR body
+    pr_body_substrings: ["==MERGE_WHEN_READY=="]
+
   # "blacklist" defines how to exclude PRs from evaluation and merging
   blacklist:
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ merge:
     # "labels" is a list of labels that must be matched to whitelist a PR for merging (case-insensitive)
     labels: ["merge when ready"]
 
-    # "comment_substrings" matches on substrings in comments
+    # "comment_substrings" matches on substrings in comments or the pull request body
     comment_substrings: ["==MERGE_WHEN_READY=="]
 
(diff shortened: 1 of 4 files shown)


bluekeyes:
Looks like this changes the behavior of `comment_substrings` to now ignore the PR body, which is technically a breaking change. I think it makes sense to have `comment_substrings` keep the old behavior of matching everything and then people that want to only match PR bodies leave that option out. I can't think of a use case right now where you'd want to allow a match _only_ in comments and not in the PR body.

asvoboda:
Sounds good. I've reverted my change and added more test cases. This is no longer a breaking change and is only additive.

README.md
```diff
@@ -36,6 +36,9 @@ merge:
     # "comment_substrings" matches on substrings in comments
```
bluekeyes: let's clarify this to `matches on substrings in comments or the pull request body`

bluekeyes (approved):

PR #72, opened 2018-11-28
Add ability to take part of pull-request body as pr description instead of all or nothing config

robert3005:
Previously bulldozer only supported summarize_commits mode or pull_request_body (with a caveat that you had to specify ==SQUASH_MSG==/==COMMIT_MSG== for the description part). Currently there's no clear alternative to the second mode. More importantly the feature I am after (and happy to change implementation as you think is appropriate) is to allow part of the pr description to be the commit description. This would be very useful for cases of auto generated prs that might have long descriptions but should be summarized in the git log in more succinct way (this is particularly annoying with github where it tracks mentions in commit descriptions for any commit interaction leading to a lot of notifications for creators of automated checks). 

Hence, I propose we standardise the feature we had previously and for pull_request_body mode add ability to select part of the pr description delineated by commit_description_delimiter paramter values.

Let me know what you think.

cc @ferozco

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/72)

### DIFF
--- a/bulldozer/config_fetcher.go
+++ b/bulldozer/config_fetcher.go
@@ -25,6 +25,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var commitMsgString = "==COMMIT_MSG=="
+
 type FetchedConfig struct {
 	Owner  string
 	Repo   string
@@ -176,7 +178,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {SummarizeCommits},
+					configv0.Strategy: {SummarizeCommits, nil},
 				},
 			},
 		}
@@ -195,7 +197,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {SummarizeCommits},
+					configv0.Strategy: {SummarizeCommits, nil},
 				},
 			},
 		}
@@ -214,7 +216,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {PullRequestBody},
+					configv0.Strategy: {PullRequestBody, &commitMsgString},
 				},
 			},
 		}
--- a/bulldozer/config_fetcher.go
+++ b/bulldozer/config_fetcher.go
@@ -178,7 +178,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {SummarizeCommits, nil},
+					configv0.Strategy: {Body: SummarizeCommits},
 				},
 			},
 		}
@@ -197,7 +197,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {SummarizeCommits, nil},
+					configv0.Strategy: {Body: SummarizeCommits},
 				},
 			},
 		}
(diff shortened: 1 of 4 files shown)


bulldozer/config_fetcher.go
```diff
@@ -25,6 +25,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var commitMsgString = "==COMMIT_MSG=="
```
bluekeyes: Since this is only used for backwards compatibility, let's inline it in the one place it is used. That will be possible once the option is not a pointer (see other comments).

bulldozer/config_fetcher.go
```diff
@@ -176,7 +178,7 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				DeleteAfterMerge: configv0.DeleteAfterMerge,
 				Method:           configv0.Strategy,
 				Options: map[MergeMethod]MergeOption{
-					configv0.Strategy: {SummarizeCommits},
+					configv0.Strategy: {SummarizeCommits, nil},
```
bluekeyes: By naming the fields, you can avoid passing `nil` or the empty string:

```go
configv0.Strategy: {Body: SummarizeCommits},
```

bulldozer/config_v1.go
```diff
@@ -53,7 +53,8 @@ type MergeConfig struct {
 }
 
 type MergeOption struct {
-	Body MessageStrategy `yaml:"body"`
+	Body                       MessageStrategy `yaml:"body"`
+	CommitDescriptionDelimiter *string         `yaml:"commit_description_delimiter,optional"`
```
bluekeyes: I'd call this `MessageDelimiter` (`message_delimiter` in YAML), make it a plain `string` (so it will be the empty string if unset), and remove the `optional` from the `yaml` tag.

bulldozer/merge.go
```diff
@@ -56,7 +57,18 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
 			if err != nil {
 				return errors.Wrap(err, "failed to determine pull request body")
 			}
-			commitMessage = body
+
+			if opt.CommitDescriptionDelimiter != nil {
+				var rString = fmt.Sprintf(
+					`(?sm:(%s\s*)^(.*)$(\s*%s))`, *opt.CommitDescriptionDelimiter, *opt.CommitDescriptionDelimiter)
```
bluekeyes: We should call `regexp.QuoteMeta` on these strings to avoid unexpected behavior

bulldozer/merge.go
```diff
@@ -56,7 +57,18 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
 			if err != nil {
 				return errors.Wrap(err, "failed to determine pull request body")
 			}
-			commitMessage = body
+
+			if opt.CommitDescriptionDelimiter != nil {
```
bluekeyes: Once this is a value instead of a pointer, compare with the empty string.

bulldozer/merge.go
```diff
@@ -56,7 +57,18 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
...
+				var rString = fmt.Sprintf(
+					`(?sm:(%s\s*)^(.*)$(\s*%s))`, *opt.CommitDescriptionDelimiter, *opt.CommitDescriptionDelimiter)
+				if m := regexp.MustCompile(rString).FindStringSubmatch(body); len(m) == 4 {
+					commitMessage = m[2]
+				} else {
+					commitMessage = body
+				}
+			} else {
+				commitMessage = body
```
bluekeyes: You can simplify this conditional by setting `commitMessage` to `body` initially and then overwriting the value if a delimiter is set and the regexp matches.

bluekeyes (commented):
This approach and feature makes sense to me, comments are about implementation details. Can you also document the new option in the README?

bulldozer/config_v1.go
robert3005: oh, I didn't know that primitives are allowed to be unset, good to know

bulldozer/merge.go
robert3005: yeah, this is one part of go that I am not fond of and always code functionally. Will change

bluekeyes (approved):

bluekeyes:
@ajlake @jmcampanini @asvoboda can I get a second check on this before we merge?

bulldozer/merge.go
```diff
@@ -56,7 +57,15 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
 			if err != nil {
 				return errors.Wrap(err, "failed to determine pull request body")
 			}
+
 			commitMessage = body
+			if opt.MessageDelimiter != "" {
```
asvoboda: I think it might be worthwhile to refactor the calculation of `commitMessage` so that its more simple to write some tests around the expected behaviour. Perhaps refactoring something like L45-L79:
```
func calculateCommitMessage() (string) {
  commitMessage := ""
   if mergeConfig.Method == SquashAndMerge {

  return commitMessage
}
```

The only bit that would worry me is the inclusion of the `MustCompile` on L65 that will panic if the regex doesn't compile correctly, so that would be where I would want to focus on for testing. Maybe we can do that in another PR, or here.

robert3005:
Yeah, we should refactor it and account for regex compilation failure here.
This code is a bit too optimistic

bluekeyes (approved):
I think this looks good. If you can add the `message_delimiter` option to the configuration example in the README, I'll merge it.

robert3005:
Updated the readme with examples

PR #75, opened 2018-11-30
Use default log level if none is specified

bluekeyes:
Otherwise, omitting the log level from the configuration disables
logging completely.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/75)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -96,13 +96,17 @@ func configureLogger(c LoggingConfig) (zerolog.Logger, error) {
 		out = zerolog.ConsoleWriter{Out: out}
 	}
 
-	logLevel, err := zerolog.ParseLevel(c.Level)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


asvoboda (approved):

server/server.go
```diff
@@ -96,13 +96,17 @@ func configureLogger(c LoggingConfig) (zerolog.Logger, error) {
 		out = zerolog.ConsoleWriter{Out: out}
 	}
 
-	logLevel, err := zerolog.ParseLevel(c.Level)
+	logger := zerolog.New(out).With().Timestamp().Logger()
+	if c.Level == "" {
```
jmcampanini, line 100: is there a way to get this into go-baseapp? i see we take a pre-built logger in when constructing a server, and we determine the writer, just feels like there is something we can do to make this less error-prone

PR #77, opened 2018-12-04
Add: repository default config support

### LINKED ISSUE #76: Feature: fallback to global configuration when repo config file not found.
Currently, we need to add a config file (eg .bulldozer.yml) to each repo.

I've add support for global configuration as repo fallback configuration. 

We can install bulldozer as GitHub app only to projects need to enable this feature.

You can view my implementation here: https://github.com/richard1122/bulldozer/commit/ee5a0b79d0855b83821cadec3eb5a73c31b05b36

If this feature is acceptable, I 'd like to send a PR.

richard1122:
background: https://github.com/palantir/bulldozer/issues/76
fix: #76 

This is my first time coding in Golang, please feel free to leave any comments.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/77)

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,20 @@ merge:
 Anything that's contained between two `==COMMIT_MSG==` strings will become the
 commit message instead of whole pull request body.
 
+#### What if I don't want to put config files into each repo?
+You can enable global fallback configuration in your bulldozer config file.
+
+Global configuration will be used only when your repo config file not exists.
+
+```yaml
+options:
+  enable_global_config: true
+
+global:
+  blacklist:
+    labels: ["do not merge"] # or any other available config.
+```
+
 #### Bulldozer isn't merging my commit or updating my branch when it should, what could be happening?
 
 Bulldozer will attempt to merge a branch whenever it passes the whitelist/blacklist
--- a/README.md
+++ b/README.md
@@ -185,15 +185,14 @@ commit message instead of whole pull request body.
 #### What if I don't want to put config files into each repo?
 You can enable global fallback configuration in your bulldozer config file.
 
-Global configuration will be used only when your repo config file not exists.
+Global configuration will be used only when your repo config file does not exist.
 
 ```yaml
 options:
   enable_global_config: true
-
-global:
-  blacklist:
-    labels: ["do not merge"] # or any other available config.
+  default_repository_config:
+    blacklist:
+      labels: ["do not merge"] # or any other available config.
 ```
 
 #### Bulldozer isn't merging my commit or updating my branch when it should, what could be happening?
--- a/README.md
+++ b/README.md
@@ -183,13 +183,12 @@ Anything that's contained between two `==COMMIT_MSG==` strings will become the
 commit message instead of whole pull request body.
 
 #### What if I don't want to put config files into each repo?
-You can enable global fallback configuration in your bulldozer config file.
+You can add default repository configuration in your bulldozer config file.
 
-Global configuration will be used only when your repo config file does not exist.
+It will be used only when your repo config file does not exist.
 
 ```yaml
 options:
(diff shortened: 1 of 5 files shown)


bulldozer/config_fetcher.go
```diff
@@ -52,12 +52,16 @@ func (fc FetchedConfig) String() string {
 type ConfigFetcher struct {
 	configurationV1Path  string
 	configurationV0Paths []string
+	enableGlobalConfig   bool
+	globalConfig         Config
```
bluekeyes: I think this could be simpler if `globalConfig` was a pointer in all places where it is used. You can then check if `globalConfig != nil` to determine if it is enabled instead of passing a second flag and requiring users to set both.

server/config.go
```diff
@@ -33,6 +35,7 @@ type Config struct {
 	Options Options            `yaml:"options"`
 	Logging LoggingConfig      `yaml:"logging"`
 	Datadog datadog.Config     `yaml:"datadog"`
+	Global  bulldozer.Config   `yaml:"global"`
```
bluekeyes: To make the configuration clearer, I'd do the following:

1. Make this a field of the `Options` struct instead of a top-level field
2. Rename it to `DefaultRepositoryConfig` (`default_repository_config` in YAML)

I think that will make it clear that this is configuration used if a repository doesn't have any, avoiding confusion with other "global" settings.

README.md
```diff
@@ -89,6 +89,20 @@ merge:
       body: summarize_commits # or `pull_request_body`, `empty_body`
 ```
 
+#### What if I don't want to put config files into each repo?
+You can enable global fallback configuration in your bulldozer config file.
+
+Global configuration will be used only when your repo config file not exists.
```
bluekeyes: style: `only when you repo config file does not exist.`

bluekeyes (commented):
Thanks for implementing this. I had a few comments on how the configuration is defined that I think will make it simpler and easier to understand.

README.md
```diff
@@ -182,6 +182,19 @@ merge:
...
 commit message instead of whole pull request body.
 
+#### What if I don't want to put config files into each repo?
+You can enable global fallback configuration in your bulldozer config file.
+
+Global configuration will be used only when your repo config file does not exist.
+
+```yaml
+options:
+  enable_global_config: true
```
richard1122: Should we rename this to `enable_repository_default_config` ?

richard1122:
Hi @bluekeyes , sorry for the late reply.
I've updated this PR and could you please have a look?

server/config.go
```diff
@@ -41,9 +43,11 @@ type LoggingConfig struct {
 }
 
 type Options struct {
-	AppName              string   `yaml:"app_name"`
-	ConfigurationPath    string   `yaml:"configuration_path"`
-	ConfigurationV0Paths []string `yaml:"configuration_v0_paths"`
+	AppName                 string           `yaml:"app_name"`
+	ConfigurationPath       string           `yaml:"configuration_path"`
+	EnableGlobalConfig      bool             `yaml:"enable_global_config"`
+	DefaultRepositoryConfig bulldozer.Config `yaml:"default_repository_config"`
```
bluekeyes: If you define this as `*bulldozer.Config`, then I think you can remove the `EnableGlobalConfig` flag and make it so that the existence of this options enables the default fallback. That will also simplify the code in `server.go` a bit.

README.md
bluekeyes: I actually think we only need the `default_repository_config` field, see my other comment.

bluekeyes (approved):
Thanks for implementing this! I pushed one commit to your branch to fix a minor thing with usage of pointers, but otherwise this looks good now.

Issue #70, opened 2018-11-26
FR: post a PR status when enabled, showing if its ready


jmcampanini:
right now, a user can't see if bulldozer is enabled on a repo.

if we post a status that:

* says pending if bulldozer is waiting for something before it can try to merge (like a whitelist tag)
* says ready if bulldozer is ready to try to merge

PR #99, opened 2019-03-21
Support target branches in whitelist/blacklist

### LINKED ISSUE #86: Whitelist/Blacklist based on target branch
Hi,

Thanks for releasing this, it's almost exactly what I need. One workflow I have seems to require a workaround, which is whitelisting/blacklisting based on the target branch of the PR. For context, we want PR's into the production environment not be to squashed, but all other PR's to be.

I've dealt with this by whitelisting labels and only applying the label to branches that should be dealt with, but it would be helpful to be able to whitelist target branches OR if it was possible to apply merge strategies based on the target branch?

I think another workaround I could apply would be to run two bulldozer apps side-by-side, but that's also a pretty big burden.

If it's something you're open to, I'd be happy to take a shot at developing the feature.

Either way, thanks for the tool :+1:

bluekeyes:
Set the "branch" option to whitelist/blacklist PRs that target any branches in the list. Also refactor the whitelist/blacklist handling to reduce code duplication.

Fixes #86.


asvoboda (approved):
👍 for deduping the blacklist/whitelist signal match.

README.md
```diff
@@ -93,6 +93,9 @@ merge:
     # to the whitelist.
     pr_body_substrings: ["==MERGE_WHEN_READY=="]
 
+    # Pull requests targeting any of these branches are added to the whitelist.
+    branches: ["develop"]
```
jmcampanini, line 98: this is cool!

non-blocking, for another PR: what's your thoughts on bulldozer providing signal that it will do merge when the checks are ready? i think it may be useful to have comms coming from bulldozer, especially when we start making what is whitelisted more implicit

bulldozer/signals.go
```diff
@@ -0,0 +1,110 @@
...
+	size += len(s.PRBodySubstrings)
+	size += len(s.Branches)
+	return size > 0
+}
+
+// Matches returns true if the pull request meets one or more signals. It also
+// returns a description of the signal that was met. The mode argument appears
+// in this description and indicates the mode (whitelist, blacklist) this set
+// of signals is associated with.
+func (s *Signals) Matches(ctx context.Context, pullCtx pull.Context, mode string) (bool, string, error) {
```
jmcampanini: nit: `mode` as a variable name makes me think it causes different behavior. i can't come up with anything better: `modeDescription` just feels bad. the godoc should be sufficient. i just dont want to end up using the mode to drive behavior

bulldozer/evaluate.go
```diff
@@ -28,130 +27,23 @@ import (
...
 func IsPRWhitelisted(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
-	labels, err := pullCtx.Labels(ctx)
+	matches, reason, err := config.Matches(ctx, pullCtx, "whitelist")
 	if err != nil {
-		return false, "unable to list PR labels", err
-	}
-
-	if inSlice, idx := anyInSliceCaseInsensitive(labels, config.Labels); inSlice {
-		return true, fmt.Sprintf("PR label matches one of specified whitelist labels: %q", config.Labels[idx]), nil
+		// whitelist must always fail closed (no match on error)
```
jmcampanini, line 43: curious: any way to enforce this via test? (same for blacklist)

jmcampanini (approved):

README.md
bluekeyes: See #70. I think that makes sense, but I'm not sure what mechanism to use to communicate that (status check or something else).

bulldozer/evaluate.go
bluekeyes: Yeah, there are already tests for this: https://github.com/palantir/bulldozer/blob/bkeyes/branch-whitelist/bulldozer/evaluate_test.go#L48

bulldozer/signals.go
bluekeyes: How about `label`?

jmcampanini: much like label

bluekeyes: Turns out `label` shadows the usage of PR labels in this function, so you're getting `tag` instead, which I think works the same.

PR #101, opened 2019-03-21
Support strategies for generating commit titles

### LINKED ISSUE #93: bulldozer commit messages are inconsistent
When a PR containing multiple commits is merged (using squash+merge), bulldozer uses the PR title as the commit message on develop.

When a PR containing a _single_ commit is merged (using squash+merge), the commit message of that single commit appears on develop, not the PR title. This is inconsistent.

It's particularly annoying because we enforce a convention of prefixing PRs with `[fix]` `[improvement]` etc in some of our OSS repos, which then get discarded. 

Example: https://github.com/palantir/conjure-java/commit/8bf6a57b5883e95ed52737c087b4bb453b023f20

bluekeyes:
When squashing commits, users can now set how the commit title is generated, since GitHub creates different titles based on the number of commits to merge. The `pull_request_title` strategy uses the title of the pull request, the `first_commit_title` strategy uses the first line of the oldest commit in the pull request, and the `github_default_title` strategy defers to GitHub's logic. The title always end with the PR number in parentheses.

Note: this PR includes two default behavior changes:

1. The default title strategy is now `pull_request_title`
2. The default message strategy is now `summarize_commits`

Previously, they were `github_default_title` and `empty_body`. I think the new defaults are more user-friendly, but they are different.

This also does some refactoring, moving commit listing to `pull.Context` and clarifying that merge options only apply to squashes.

Fixes #93. This also deserves some tests, which I'll add in another commit before merging.


README.md
```diff
@@ -120,12 +120,18 @@ merge:
   options:
     # "squash" options are only used when the merge method is "squash"
     squash:
+      # "title" defines how the title of the commits message is created when
+      # generating a squash commit. The options are "pull_request_title",
+      # "first_commit_title", and "github_default_title".
+      title: "pull_request_title"
```
jmcampanini, line 127: do we mention that `pull_request_title` and `summarize_commits` are defaults anywhere in the README?

bulldozer/merge.go
```diff
@@ -187,51 +200,53 @@ func calculateCommitMessage(ctx context.Context, pullCtx pull.Context, client *g
...
+	case FirstCommitTitle:
+		commits, err := pullCtx.Commits(ctx)
+		if err != nil {
+			return "", err
+		}
+		// TODO(bkeyes): check this assumption
+		// commits are ordered from newest to oldest
+		firstCommit := commits[len(commits)-1]
+		title = strings.SplitN(firstCommit.Message, "\n", 1)[0]
+	case GithubDefaultTitle:
```
jmcampanini, line 230: if default, it'll return an empty string? sometimes i do wish there was an optional that wasnt null

pull/github_context.go
```diff
@@ -115,6 +115,35 @@ func (ghc *GithubContext) Comments(ctx context.Context) ([]string, error) {
...
+	}
+
+	commits := make([]*Commit, len(allCommits))
+	for i, c := range allCommits {
+		commits[i] = &Commit{
+			SHA:     c.GetCommit().GetSHA(),
+			Message: c.GetCommit().GetMessage(),
+		}
+	}
+	return commits, nil
```
jmcampanini: should we cache this?

jmcampanini (approved):
looking good! only question i have is about considering the change in default behavior a breaking change or not. i dont mind either answer, just figure it should be on purpose

bulldozer/merge.go
bluekeyes: Yeah, it will return the default value of `title`, which is empty, which GitHub will then interpret as the default/unset behavior.

pull/github_context.go
bluekeyes: Good catch, I think the other code I wrote assumed this was cached.

bluekeyes:
Based on #93, it seems like people already expect the new behavior in this PR for titles and therefore it is friendlier to consider the old behavior a bug and change the default, rather than requiring updates to configuration files everywhere. But thinking about it now, I don't see a reason to change the default message behavior to `summarize_commits` - I'll leave that as `empty_body`.

I'll call the title change out in the release notes and the README, but I don't think it merits a new major version number.

README.md
bluekeyes: Add a line about the default values.

iamdanfox:
Thanks @bluekeyes :D

Issue #98, opened 2019-03-15
Unable to Merge when GitHub Branch Protection Push Restrictions Enabled


BrianChristie:
Is there any current workarounds to use Bulldozer when `Restrict who can push to matching branches` is enabled on GitHub?

From looking at other projects, it appears it's currently impossible as GitHub does not provide a way to give push permission to the Bulldozer GitHub App directly, only user accounts or groups can be selected.

Some projects have worked around this by having a 'real' user account for the bot to use in this situation.

- Same issue in Renovate: https://github.com/renovatebot/renovate/issues/846
- Similar issue in dependabot: https://github.com/dependabot/feedback/issues/86

Thanks!

bluekeyes:
Internally, we recommend avoiding push restrictions for this reason. We use [policy-bot](https://github.com/palantir/policy-bot) to implement approval restrictions as a required status check, which solves most of the problems we'd otherwise use push restrictions to solve. But this isn't a great solution for everyone, as it requires running a second app and doesn't enforce exactly the same thing.

I don't know of any other workaround at the moment (and I'd really like GitHub to just allow bot users to exist in that restrictions list), but adding an option to configure an OAuth token to use for merges sounds like an acceptable feature until that day. For simplicity, we'd probably use that token for all merges if it is configured, rather than only using it if the repository has push restrictions.

PR #105, opened 2019-03-25
Refactor GitHub client usage for merges

### LINKED ISSUE #100: Improve how pull.GithubContext uses PR information
`pull.GithubContext` is a bit weird about how it uses `github.PullRequest` objects:

1. It takes one in the constructor and several functions assume it is non-nil, but the constructor also takes 3 parameters that duplicate information from the PR
2. `Body()` just returns the body value of the saved PR object, but takes a context and returns an error
3. `Title()` just returns the title value of the saved PR object, but takes a context and returns an error

Assuming all callers have a full (or full-enough) `github.PullRequest`, we should only require that as an argument and then simplify any methods that only use data from the PR and make no additional API calls.

bluekeyes:
Add additional methods to `pull.Context` (and clean up some of the existing ones) so that `bulldozer.MergePR` can operate without any direct references to a `github.Client`. This improves testability (although I don't add any new tests here) and enables a fix for #98 that I'll add in a followup PR.

Fixes #100.

### DIFF
--- a/bulldozer/merge.go
+++ b/bulldozer/merge.go
@@ -29,9 +29,42 @@ import (
 	"github.com/palantir/bulldozer/pull"
 )
 
+type Merger interface {
+	// Merge merges the pull request in the context using the commit message
+	// and options. It returns the SHA of the merge commit on success.
+	Merge(ctx context.Context, pullCtx pull.Context, message string, options *github.PullRequestOptions) (string, error)
+
+	// DeleteHead deletes the head branch of the pull request in the context.
+	DeleteHead(ctx context.Context, pullCtx pull.Context) error
+}
+
+type GitHubMerger struct {
+	client *github.Client
+}
+
+func NewGitHubMerger(client *github.Client) Merger {
+	return &GitHubMerger{
+		client: client,
+	}
+}
+
+func (m *GitHubMerger) Merge(ctx context.Context, pullCtx pull.Context, message string, options *github.PullRequestOptions) (string, error) {
+	result, _, err := m.client.PullRequests.Merge(ctx, pullCtx.Owner(), pullCtx.Repo(), pullCtx.Number(), message, options)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	return result.GetSHA(), nil
+}
+
+func (m *GitHubMerger) DeleteHead(ctx context.Context, pullCtx pull.Context) error {
+	_, head := pullCtx.Branches()
+	_, err := m.client.Git.DeleteRef(ctx, pullCtx.Owner(), pullCtx.Repo(), fmt.Sprintf("refs/heads/%s", head))
+	return errors.WithStack(err)
+}
+
 const MaxPullRequestPollCount = 5
 
-func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, mergeConfig MergeConfig) error {
+func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConfig MergeConfig) error {
 	logger := zerolog.Ctx(ctx)
 
 	mergeOpts := &github.PullRequestOptions{}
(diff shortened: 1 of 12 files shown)


bulldozer/merge.go
```diff
@@ -89,78 +117,75 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
...
 				if mergeConfig.DeleteAfterMerge {
-					ref := fmt.Sprintf("refs/heads/%s", pr.Head.GetRef())
+					ref := fmt.Sprintf("refs/heads/%s", head)
 
 					// check other open PRs to make sure that nothing is trying to merge into the ref we're about to delete
-					prs, err := pull.ListOpenPullRequestsForRef(ctx, client, pullCtx.Owner(), pullCtx.Repo(), ref)
+					isTargeted, err := pullCtx.IsTargeted(ctx)
 					if err != nil {
-						logger.Error().Err(errors.WithStack(err)).Msgf("Unable to list open prs against ref %s to compare delete request", ref)
+						logger.Error().Err(err).Msgf("Unable to determine if ref %s is targeted by other open pull requests before deletion", ref)
```
asvoboda, line 179: nit: Why the change away from `errors.WithStack(err)`?

pull/github_context.go
```diff
@@ -67,12 +67,24 @@ func (ghc *GithubContext) Locator() string {
 	return fmt.Sprintf("%s/%s#%d", ghc.owner, ghc.repo, ghc.number)
 }
 
-func (ghc *GithubContext) Title(ctx context.Context) (string, error) {
-	return ghc.pr.GetTitle(), nil
```
asvoboda, line 71: 🥇 Huh. How did this pass in CR previously..

bulldozer/merge.go
```diff
@@ -89,78 +117,75 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
...
 				}
 			}
 
-			logger.Info().Msgf("Successfully merged pull request for sha %s with message %q", result.GetSHA(), result.GetMessage())
+			logger.Info().Msgf("Successfully merged pull request as SHA %s", sha)
 
-			// Delete ref if owner of BASE and HEAD match
-			// otherwise, its from a fork that we cannot delete
-			if pr.GetBase().GetUser().GetLogin() == pr.GetHead().GetUser().GetLogin() {
+			// if head is qualified (contains ":"), PR is from a fork and we don't have delete permission
```
asvoboda, line 171: I think we should address this in a later PR, but if the remote fork has bulldozer installed this assumption is no longer correct. I think it was originally here as a workaround from when this operated using oauth tokens. 

We might be able to attempt a delete action, and just fail if it doesn't work. Not blocking, though.

bulldozer/merge.go
```diff
@@ -181,12 +206,7 @@ func calculateCommitMessage(ctx context.Context, pullCtx pull.Context, option Sq
 	commitMessage := ""
 	switch option.Body {
 	case PullRequestBody:
-		body, err := pullCtx.Body(ctx)
-		if err != nil {
-			return "", err
-		}
-
-		commitMessage = body
+		commitMessage = pullCtx.Body()
```
asvoboda, line 214: 🎉

asvoboda (approved):
+1 to the additional cleanup, as well.

bulldozer/merge.go
```diff
@@ -29,19 +29,47 @@ import (
 	"github.com/palantir/bulldozer/pull"
 )
 
-const MaxPullRequestPollCount = 5
+type Merger interface {
+	// Merge merges the pull request in the context using the commit message
+	// and options. It returns the SHA of the merge commit on success.
+	Merge(ctx context.Context, pullCtx pull.Context, message string, options *github.PullRequestOptions) (string, error)
```
jmcampanini: non-blocking: is it worth also abstracting out the `PullRequestOptions`, so Merger can have no ties to the `github` package? it feels like busy work to do that, but wanted to ask

bulldozer/merge.go
```diff
@@ -89,78 +117,75 @@ func MergePR(ctx context.Context, pullCtx pull.Context, client *github.Client, m
...
-					logger.Info().Msgf("Merge rejected due to unsatisfied condition %q", gerr.Message)
+					logger.Info().Msgf("Merge rejected due to unsatisfied condition: %q", gerr.Message)
 					return
 				case http.StatusConflict:
-					logger.Info().Msgf("Merge rejected due to being invalid %q", gerr.Message)
+					logger.Info().Msgf("Merge rejected due to being invalid: %q", gerr.Message)
 					return
 				default:
-					logger.Error().Err(errors.WithStack(err)).Msgf("Merge failed unexpectedly %q", gerr.Message)
+					logger.Error().Msgf("Merge failed with unexpected status: %d: %q", gerr.Response.StatusCode, gerr.Message)
```
jmcampanini, line 164: losing the stacktrace on purpose?

jmcampanini (approved):

bulldozer/merge.go
bluekeyes: I think that makes sense, mostly because it would remove the split between the commit message and the other options.

bulldozer/merge.go
bluekeyes: Yeah, we already know that the error in this case is an HTTP failure, so the important part is the status code and the message. It's also unambiguous where the error is generated.

bulldozer/merge.go
bluekeyes: If we want to support this, I consider it a new feature that requires more refactoring: even if the fork has Bulldozer installed, we need to look up the installation ID and construct a new client using that ID.

I'm also not sure how useful this would be. It seems odd to need Bulldozer for merges on a repository that mostly submits PRs to an upstream repo, so would the only reason to install it be so that it could delete branches?

bulldozer/merge.go
bluekeyes: I assume that `pull.Context` implementations already wrap errors as needed, so adding a stack again is redundant. It was useful before because the code was mostly dealing with the GitHub client directly.

PR #106, opened 2019-03-26
Add support for merging to restricted branches

### LINKED ISSUE #98: Unable to Merge when GitHub Branch Protection Push Restrictions Enabled

bluekeyes:
Users can configure a personal access token for a normal user and Bulldozer will use that token to merge into protected branches with push restrictions. This configuration is optional and only used when push restrictions are active.

Fixes #98 and requires #105 to merge first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ Anything that's contained between two `==COMMIT_MSG==` strings will become the
 commit message instead of whole pull request body.
 
 #### What if I don't want to put config files into each repo?
+
 You can add default repository configuration in your bulldozer config file.
 
 It will be used only when your repo config file does not exist.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -214,9 +215,38 @@ which are to be expected, and others that may be caused by mis-configuring Bulld
...
+   requiring changes to go through the pull request process and be approved.
+
+2. Configure Bulldozer to use a personal access token for a regular user to
+   perform merges in this case. The token must have the `repo` scope and the
+   user must be allowed to push to the branch. In the server configuration
+   file, set:
+
+   ```yaml
+   options:
+     restriction_user_token: <token-value>
```
asvoboda: nit: I think the name should  include `push` to match the existing struct's name: `PushRestrictionMerger`. Perhaps `push_restriction_user_token`?

bulldozer/merge.go
```diff
@@ -62,8 +65,49 @@ func (m *GitHubMerger) DeleteHead(ctx context.Context, pullCtx pull.Context) err
...
+func (m *PushRestrictionMerger) DeleteHead(ctx context.Context, pullCtx pull.Context) error {
+	restricted, err := pullCtx.PushRestrictions(ctx)
+	if err != nil {
+		return err
+	}
+
+	// this is not necessary: the normal client should have delete permissions,
+	// but having the merge user also delete the branch is a better UX
+	if restricted {
+		return m.Restricted.DeleteHead(ctx, pullCtx)
```
asvoboda, line 115: Do you think it would be valuable here to also log that we're performing a delete action and split it out by adding if we're using the restricted user token or not?

asvoboda (approved):

jmcampanini (approved):
lgtm

bulldozer/merge.go
bluekeyes: The action is logged by the calling code, but I'll add a message about which client was selected.

PR #108, opened 2019-04-05
[WIP] Support Github Check Suites

renefritze:
I'm trying to address #96 
As a first step I've tried to add a Handler for the `check_suite` webhook. It only differs in small details from the `status` one. Apparently I missed something in enabling the handler though, since in my test it never gets fired. I get a log like this:
> bulldozer_1  | 2019-04-05T09:24:03.657579750Z |INFO| Received webhook event github_delivery_id=a9e8ebd0-5782-11e9-8ae1-0c2e5b037836 github_event_type=check_suite rid=bijhtcqglpu77frvu2r0
bulldozer_1  | 2019-04-05T09:24:03.657636049Z |INFO| http_request elapsed=22.201649 method=POST path=/api/github/hook rid=bijhtcqglpu77frvu2r0 size=0 status=202

but nothing from the handler it seems.

PR #110, opened 2019-04-18
Add support for Github Checks

### LINKED ISSUE #96: Support status checks created by Checks API
I've deployed bulldozer via the 1.4.0 tagged docker image and setup a GithubApp for our organization. I'm now trying out the config in a toy repository. In [this PR](https://github.com/pymor/ci_playground/pull/4) the first comment is matching the configured whitelisted "comment" and the second comment the "comment_substring". In the debug output however I see

> pymor_bulldozer_1  | 2019-03-07T08:03:17.357768662Z |DEBUG| pymor/ci_playground#4 is deemed not mergeable because of unfulfilled status checks: [pymor.ci_playground] 

Which seems wrong since that is the name of the only required status check and that passed.

chdsbd:
Currently only Github status checks are supported. This change adds support for the newer Github Checks API.

This is based on https://github.com/palantir/bulldozer/pull/108.
I replaced the CheckSuite handler with a CheckRun one. CheckSuiteEvents are only sent only once per ref, whereas a CheckRun can be sent multiple times, in the case of the [WIP app](https://github.com/apps/wip).

This change requires updated permissions to allow us to lookup Checks for specific refs and to listen to CheckRunEvents.

Fixes #96

### DIFF
--- a/pull/github_context.go
+++ b/pull/github_context.go
@@ -224,9 +224,9 @@ func (ghc *GithubContext) CurrentSuccessStatuses(ctx context.Context) ([]string,
 			opts.Page = res.NextPage
 		}
 
-		CheckOpts := &github.ListCheckRunsOptions{ListOptions: github.ListOptions{PerPage: 100}}
+		checkOpts := &github.ListCheckRunsOptions{ListOptions: github.ListOptions{PerPage: 100}}
 		for {
-			checkRuns, res, err := ghc.client.Checks.ListCheckRunsForRef(ctx, ghc.owner, ghc.repo, ghc.pr.GetHead().GetSHA(), CheckOpts)
+			checkRuns, res, err := ghc.client.Checks.ListCheckRunsForRef(ctx, ghc.owner, ghc.repo, ghc.pr.GetHead().GetSHA(), checkOpts)
 			if err != nil {
 				return ghc.successStatuses, errors.Wrapf(err, "cannot get check runs for SHA %s on %s", ghc.pr.GetHead().GetSHA(), ghc.Locator())
 			}
@@ -240,7 +240,7 @@ func (ghc *GithubContext) CurrentSuccessStatuses(ctx context.Context) ([]string,
 			if res.NextPage == 0 {
 				break
 			}
-			CheckOpts.Page = res.NextPage
+			checkOpts.Page = res.NextPage
 		}
 
 		ghc.successStatuses = successStatuses
--- a/pull/github_context.go
+++ b/pull/github_context.go
@@ -224,6 +224,25 @@ func (ghc *GithubContext) CurrentSuccessStatuses(ctx context.Context) ([]string,
 			opts.Page = res.NextPage
 		}
 
+		check_opts := &github.ListCheckRunsOptions{ListOptions: github.ListOptions{PerPage: 100}}
+		for {
+			checkRuns, res, err := ghc.client.Checks.ListCheckRunsForRef(ctx, ghc.owner, ghc.repo, ghc.pr.GetHead().GetSHA(), check_opts)
+			if err != nil {
+				return ghc.successStatuses, errors.Wrapf(err, "cannot get check runs for SHA %s on %s", ghc.pr.GetHead().GetSHA(), ghc.Locator())
+			}
+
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @chdsbd! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pull/github_context.go
```diff
@@ -224,6 +224,25 @@ func (ghc *GithubContext) CurrentSuccessStatuses(ctx context.Context) ([]string,
 			opts.Page = res.NextPage
 		}
 
+		checkOpts := &github.ListCheckRunsOptions{ListOptions: github.ListOptions{PerPage: 100}}
+		for {
+			checkRuns, res, err := ghc.client.Checks.ListCheckRunsForRef(ctx, ghc.owner, ghc.repo, ghc.pr.GetHead().GetSHA(), checkOpts)
```
bluekeyes, line 229: Does this returns an empty list when there are no check runs for the ref or a 404? I know there are some GitHub APIs that behave one way and some that behave the other.

bluekeyes (commented):
Thanks for implementing this! The overall approach looks good, so I'll try to do a more detailed review in the next few days after reading a bit more about the Checks API.

pull/github_context.go
chdsbd: Looks like it returns an empty array.

```console
> curl -H 'Accept: application/vnd.github.antiope-preview+json' 'https://api.github.com/repos/palantir/bulldozer/commits/08efe6866fe81693f6da541cc4575a49444b656a/check-runs?per_page=100'
{
  "total_count": 0,
  "check_runs": [

  ]
}
```

server/handler/check_run.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (h *CheckRun) Handles() []string {
+	return []string{"check_run"}
+}
+
+func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, payload []byte) error {
+	var event github.CheckRunEvent
+
+	if err := json.Unmarshal(payload, &event); err != nil {
+		return errors.Wrap(err, "failed to parse status event payload")
```
bluekeyes: typo: `parse check_run event payload`

server/handler/check_run.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	repo := event.GetRepo()
+	installationID := githubapp.GetInstallationIDFromEvent(&event)
+
+	ctx, logger := githubapp.PrepareRepoContext(ctx, installationID, repo)
+	logger.Debug().Msgf("Doing nothing since check_run action was %q instead of 'completed'", event.GetAction())
```
bluekeyes: I think we only want the instance of this message that appears on L50.

server/handler/check_run.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if event.GetAction() != "completed" {
+		logger.Debug().Msgf("Doing nothing since check_run action was %q instead of 'completed'", event.GetAction())
+		return nil
+	}
+	suite := event.GetCheckRun()
```
bluekeyes: Should this be `run` instead of `suite`?

bluekeyes (changes requested):
Looks good overall, just a few minor issues to fix before merging.

server/handler/check_run.go
chdsbd: It was my original understanding that CheckRuns were needed because CheckSuites would only be sent once per commit, but now I'm not sure about that. I think both event types would work.

bluekeyes (approved):

server/handler/check_run.go
bluekeyes: Sorry, I wasn't clear - I only meant the variable name didn't match the method call. From my reading, I think `check_run` is the correct event.

bluekeyes:
The `palantirtechnologies/bulldozer:snapshot` Docker image was updated with this if you want to test it out. I'll tag a new release after some additional internal testing.

PR #123, opened 2019-06-10
Use in memory http caching

asvoboda:


.gitignore
```diff
@@ -3,5 +3,6 @@
 *.iml
 *.ipr
 *.iws
+bulldozer.yml
```
jmcampanini: curious: why this?

asvoboda: I can remove this. I was using this path for iterating locally with the configuration.

jmcampanini: all good. not sure if we should leave this, or encourage the use of the `config/*` folder for holding configs even for use while dev-ing. was just wondering cause it looked a bit unrelated

server/server.go
```diff
@@ -53,6 +54,7 @@ func New(c *Config) (*Server, error) {
 		c.Github,
 		githubapp.WithClientUserAgent(userAgent),
 		githubapp.WithClientMiddleware(
+			githubapp.ClientCaching(func() httpcache.Cache { return httpcache.NewMemoryCache() }),
```
bluekeyes, line 57: We'll have to watch the memory usage of this after deployment. I'm not sure how much memory will actually be used by cached requests over the lifetime of the service, so we may eventually want something like https://github.com/die-net/lrucache instead.

.gitignore
bluekeyes: policy-bot has an ignore for `config/policy-bot.yml`, which is the default config location for the server. That looks like it is also the default configuration location for Bulldozer (`config/bulldozer.yml`), so if we're going to add an ignore, let's make it match the server defaults.

bluekeyes (approved):

PR #129, opened 2019-06-20
Use built in client http caching

asvoboda:


bluekeyes (approved):

server/server.go
```diff
@@ -52,6 +53,7 @@ func New(c *Config) (*Server, error) {
 	clientCreator, err := githubapp.NewDefaultCachingClientCreator(
 		c.Github,
 		githubapp.WithClientUserAgent(userAgent),
+		githubapp.WithClientCaching(true, func() httpcache.Cache { return httpcache.NewMemoryCache() }),
```
bluekeyes: Actually, thinking about this more, I believe we want to use the LRU implementation from start. Bulldozer's set of working objects tends to be new/active PRs and a tail of things that are waiting for review or being updated. Once a PR is merged, the application doesn't care about any state related to that PR, but all those object will stick around in the cache forever.

That means it's guaranteed that the cache will grow without bound, eventually using all memory, unless the application is restarted regularly.

bluekeyes (changes requested):

server/config.go
```diff
@@ -35,13 +36,19 @@ type Config struct {
...
 	Datadog datadog.Config     `yaml:"datadog"`
+	Cache   CacheConfig        `yaml:"cache"`
 }
 
 type LoggingConfig struct {
 	Level string `yaml:"level"`
 	Text  bool   `yaml:"text"`
 }
 
+type CacheConfig struct {
```
asvoboda, line 47: Do you think we should instead set this config struct in the githubapp library?

server/server.go
```diff
@@ -48,10 +51,22 @@ func New(c *Config) (*Server, error) {
 		return nil, errors.Wrap(err, "failed to initialize base server")
 	}
 
+	// Default to 50 MB for the cache size
+	maxSize := int64(50 * 1024 * 1024)
```
asvoboda: Is there a nice stdlib package that we can use so operators could write something like `50MB` in the config file?

asvoboda: Maybe https://godoc.org/github.com/inhies/go-bytesize?

server/config.go
bluekeyes: Let's leave this here for now, with the reasoning that we're not recommending any particular cache implementation in `go-githubapp`. If we end up using it a lot, I could see moving the config up a level and adding something like `WithClientLRUCaching(config)`, but I don't think the config struct alone is valuable without something in the library that uses it.

server/server.go
bluekeyes: Yeah, nothing in the standard library, but I'd be fine adding either the library you mentioned or one of these:

- https://godoc.org/github.com/docker/go-units
- https://godoc.org/github.com/c2h5oh/datasize

Because `c2h50h/datasize` implements `UnmarshalText`, you _might_ be able to reference it directly in the configuration struct and have the YAML library automatically parse it... not sure if `gopkg.in/yaml.v2` supports that or if it is only `encoding/json`.

server/server.go
```diff
@@ -48,10 +52,25 @@ func New(c *Config) (*Server, error) {
...
+	maxAge := 20 * time.Minute
+	if c.Cache.MaxAge != 0 {
+		maxAge = c.Cache.MaxAge
+	}
+
 	userAgent := fmt.Sprintf("%s/%s", c.Options.AppName, version.GetVersion())
 	clientCreator, err := githubapp.NewDefaultCachingClientCreator(
 		c.Github,
 		githubapp.WithClientUserAgent(userAgent),
+		githubapp.WithClientCaching(true, func() httpcache.Cache { return lrucache.New(maxSize, maxAge.Nanoseconds()) }),
```
bluekeyes: From https://godoc.org/github.com/die-net/lrucache#New:

>  If maxAge > 0, entries will also be expired after maxAge seconds.

That said, I'm more interested in the size limit than the time limit... I'm not sure if we get any benefit from time-based expiration as long as it can't use unbounded memory.

asvoboda: I'd be fine just removing that param 👍

server/server.go
asvoboda: Oh nice find on the later package, which seems to work as expected.

bluekeyes (approved):

PR #131, opened 2019-06-25
Do not use pull request body as message if delimiters are not present

pkoenig10:
#72 merged quite a while ago but nearly all excavator PR still do not use the delimiters. Thus all excavator PRs have very long commit messages that make it cumbersome to view the commit history.

This PR changes the behavior to explicitly exclude the pull request body if the PR does not include the specified delimiter. Now, by default excavator PRs will not have their body included in the commit message unless the explicitly opt-in and include the delimiters.


bluekeyes (approved):
Thanks for adding this. While it is a behavioral change, I think this is the behavior that was implied by the documentation for the `message_delimiter` option.

pkoenig10:
@bluekeyes can we cut a release with this change?

PR #139, opened 2019-07-23
Read some config values from the environment

### LINKED ISSUE #126: github secrets read from kubernetes Secret object
Hi there, I'm trying to build a helm chart from this project, but the `github.app` variables I don't want to set in a `values.yaml` file, no on the commandline. What would be really cool, is if there could be a way to read these required details from a k8s secret that contains everything you need.

It would be easy for me to mount the secrets appropriately in the pod, then the `github` section could have a `secrets_path`

bluekeyes:
All GitHub options and certain sensitive Bulldozer options can be
defined or overridden by environment variables. This also cleans up some
configuration-related content in the README and example file.

Fixes #126.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -15,6 +15,8 @@
 package server
 
 import (
+	"os"
+
 	"github.com/c2h5oh/datasize"
 	"github.com/palantir/go-baseapp/baseapp"
 	"github.com/palantir/go-baseapp/baseapp/datadog"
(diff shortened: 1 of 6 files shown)


server/config.go
```diff
@@ -57,22 +54,16 @@ type Options struct {
 	ConfigurationV0Paths []string `yaml:"configuration_v0_paths"`
 }
 
-func (o *Options) fillDefaults() {
```
asvoboda, line 60: Nice catch on the unused func

asvoboda (approved):

PR #141, opened 2019-07-29
follow doc for github default title strategy

clakech:


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @clakech! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes:
Thanks for catching this, @clakech. Since changing the configuration value is technically a breaking change, I'd prefer to update the documentation to match what the code does rather than the other way around.

clakech:
> Thanks for catching this, @clakech. Since changing the configuration value is technically a breaking change, I'd prefer to update the documentation to match what the code does rather than the other way around.

I think it is better this way.

If you update the doc according to the doc then your title strategies names will be inconsistent ;)

- pull_request_title
- first_commit_title
- github_default

PR #142, opened 2019-07-29
fix(check_run): use full pr to retrieve event infos

clakech:
As documented https://developer.github.com/v3/activity/events/types/#checkrunevent the check_run event pullrequests does not always contains base / repo / owner / login so as we already got the full pull request available in the code, we can use it to be sure to always have all the informations and be able to retrieve the bulldozer config file.

When the check_run event does not contains the pullrequest base / repo / owner / login info, then the bot cannot retrieve config info and failed to manage the event. Sometimes the check_run event is populated with pr / base / repo / owner / login, sometimes not. Still don't know why, we need to investigate more.

BTW, bulldozer should not use payload properties that are not documented to be present event if they are sometimes provided by github hooks. as documented here again: https://developer.github.com/v3/activity/events/types/#checkrunevent

Hope that helps

### DIFF
--- a/server/handler/check_run.go
+++ b/server/handler/check_run.go
@@ -73,7 +73,7 @@ func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, pay
 		pullCtx := pull.NewGithubContext(client, fullPR)
 
 		logger := logger.With().Int(githubapp.LogKeyPRNum, pr.GetNumber()).Logger()
-		if err := h.ProcessPullRequest(logger.WithContext(ctx), pullCtx, client, pr); err != nil {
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @clakech! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes (approved):
Thanks for fixing this. It looks like we forgot to change that parameter when `fullPR` was introduced (for the reasons you described about inconsistent API response objects) and as we don't use check runs internally yet, it wasn't caught.

clakech:
my pleasure, we use a forked version of this repo now, we will be back for your next release ;)

PR #148, opened 2019-08-10
PR merge can be triggered by push event

### LINKED ISSUE #147: PR merge can be triggered by push event
In the following sequence of operations PR merge should be triggered by push event, but it is not:

- Alice submits a PR `alice-feature` which conflicts with `master`
- Bob submits a PR `bob-feature` that removes the conflict with `alice-feature` from `master`
- Bob's PR gets merged ("push" event)
- ...?

misha-ridge:
In the following sequence of operations PR merge should be triggered by push
event:

- Alice submits a PR `alice-feature` which conflicts with `master`
- Bob submits a PR `bob-feature` that removes the conflict with `alice-feature`
  from `master`
- Bob's PR gets merged ("push" event)
- ...?

Fixes #147.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/bulldozer/148)

### DIFF
--- a/server/handler/push.go
+++ b/server/handler/push.go
@@ -80,6 +80,10 @@ func (h *Push) Handle(ctx context.Context, eventType, deliveryID string, payload
 		if err := h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, pr, baseRef); err != nil {
 			logger.Error().Err(errors.WithStack(err)).Msg("Error updating pull request")
 		}
+
+		if err := h.ProcessPullRequest(ctx, pullCtx, client, pr); err != nil {
+			logger.Error().Err(errors.WithStack(err)).Msg("Error processing pull request")
+		}
 	}
 
 	return nil


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @misha-tectonic! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

server/handler/push.go
```diff
@@ -80,6 +80,10 @@ func (h *Push) Handle(ctx context.Context, eventType, deliveryID string, payload
 		if err := h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, pr, baseRef); err != nil {
```
bluekeyes, line 80: I think it would be a good idea to change `UpdatePullRequest` to return `(bool, error)`, with the first value indicating if an update was started or not. That way, only PRs that are _not_ updated will be processed for merging; the update process should trigger the normal hooks in the other case.

bluekeyes (commented):
Good catch, we definitely didn't consider this case. To avoid excess processing, I think we only want to process a PR if it's not being updated.

server/handler/push.go
misha-ridge: `UpdatePullRequest` only starts the update, it does not check that the branch is actually behind the base, so the following happens:

Event 1 (PR branch is behind base):
- `UpdatePullRequest` returns `true`, as PR can be updated
- Merge is not performed
- Update is performed asynchronously.

Event 2 (PR branch is up-to-date):
- `UpdatePullRequest` still returns `true`, as branch still can be updated
- Merge is not performed
- Update is skipped asynchronously, as branch is up-to-date.

The only way to sequence these operations is to wait for update to check whether it's behind the base or not. May not be a bad idea, but needs some more refactoring.

bluekeyes: That's fair, I forgot that `UpdatePullRequest` only checks if updates are enabled, not if an update will actually be made.

bluekeyes (approved):
Let's try this out as is, I think the internal conditional request logic will be enough to avoid excessive requests as a result of the additional processing.

bluekeyes:
Once the CLA is signed, I'll merge this - please let us know if you have any issues with the CLA process.

misha-ridge:
@bluekeyes Sure, I'll sign it once I hear back from the legal team.

misha-ridge:
Can't get the permission from the legal.

PR #152, opened 2019-08-14
Upgrade go-githubapp to the latest version

bluekeyes:
This includes a potential fix for a race condition in middleware
application that led to a random nil pointer exception in Bulldozer.

This might fix the problem in #149.


asvoboda:
👍

asvoboda (approved):

PR #171, opened 2020-03-05
Update guidance on restricted branches

asvoboda:


jamestoyer (approved):

bluekeyes:
We should probably deprecate the workaround token feature as well and then eventually remove it. Should simplify the code a decent amount, since there's logic to specifically detect if push restrictions are enabled.

PR #176, opened 2020-03-31
Add support for ff-only merges

dlwyatt:


bulldozer/merge.go
```diff
@@ -57,6 +57,39 @@ func NewGitHubMerger(client *github.Client) Merger {
...
+}
+
+func (m *GitHubMerger) ffOnlyMerge(ctx context.Context, pullCtx pull.Context) (string, error) {
+	// ff-only merge is accomplished by calling Git.UpdateRef with the force parameter set to false, and the new commit hash for the base branch's pointer
+
+	base, _ := pullCtx.Branches()
+
+	ref, _, err := m.client.Git.GetRef(ctx, pullCtx.Owner(), pullCtx.Repo(), fmt.Sprintf("refs/heads/%s", base))
+	if err != nil {
+		return "", err
```
bluekeyes: style: let's wrap these with `errors.Wrap` to attach stack information since they are coming from third-party code

bulldozer/merge.go
```diff
@@ -57,6 +57,39 @@ func NewGitHubMerger(client *github.Client) Merger {
...
+	headCommitSHA := pullCtx.HeadSHA()
+	ref.Object.SHA = &headCommitSHA
+
+	newRef, _, err := m.client.Git.UpdateRef(ctx, pullCtx.Owner(), pullCtx.Repo(), ref, false)
+	if err != nil {
+		return "", err
+	}
+
+	if newRef.GetObject().GetSHA() != headCommitSHA {
+		return "", fmt.Errorf("Expected reference to be updated to SHA %s, but instead it points to %s", headCommitSHA, newRef.GetObject().GetSHA())
```
bluekeyes, line 86: Is this the response when a fast-forward is not possible, i.e. GitHub returns "success", but the ref doesn't actually change? If so, it would be helpful to indicate that in this message, maybe as `could not fast-foward $oldsha to $newsha`.

bluekeyes (commented):
This is neat, I forgot you can "merge" PRs by including their code on the target branch via a ref update. Some minor style comments, but looks good otherwise.

bulldozer/merge.go
dlwyatt: honestly, I'm not sure if this situation will ever happen.  If it would return a different hash here, the call to UpdateRef _should_ return an error, but since it gives us the new reference as a return value, figured I'd check it just in case.

bulldozer/merge.go
dlwyatt: will do

dlwyatt:
@bluekeyes Any other updates needed for this?

bulldozer/merge.go
```diff
@@ -57,6 +57,39 @@ func NewGitHubMerger(client *github.Client) Merger {
...
+	if err != nil {
+		return "", errors.Wrap(err, "Could not get git reference of PR base branch")
+	}
+
+	headCommitSHA := pullCtx.HeadSHA()
+	ref.Object.SHA = &headCommitSHA
+
+	newRef, _, err := m.client.Git.UpdateRef(ctx, pullCtx.Owner(), pullCtx.Repo(), ref, false)
+	if err != nil {
+		return "", err
```
bluekeyes: Can you wrap this error as well?

bluekeyes (commented):
One more error to wrap, but looks good otherwise.

bulldozer/merge.go
dlwyatt: Done!

bluekeyes (approved):

PR #179, opened 2020-04-24
Convert to Go modules, upgrade dependencies

bluekeyes:
This upgrades to palantir/go-githubapp v0.2.0 and google/go-github
v30.1.0. I'll upgrade both of these further in a future commit but
wanted to minimize non-module changes in this one.


bluekeyes:
Note that I'm deliberately delaying on the same conversion for `policy-bot` because of some breaking API changes in `google/go-github@v29.0.3` and later. Bulldozer doesn't use the Team APIs, so there's no issue (that I know of) here.

I plan to do the `policy-bot` conversion next, but want to validate the conversion and the new asynchronous dispatch feature here first.

jamestoyer (approved):

PR #180, opened 2020-04-24
Enable asynchronous dispatch for webhooks

bluekeyes:
This involved some refactoring in the UpdatePR and MergePR methods,
which implemented their own async processing previously. As a
consequence of including the newer version of go-githubapp, this also
upgrades google/go-github to v31.

The default queue size and worker count should be more than sufficient
for most users. Looking at our internal instance, which is heavily used,
shows that the p99 value for concurrent requests rarely exceeds 16
combined between two servers. That means the per-server concurrent
request count is rarely more than 8.


bulldozer/merge.go
```diff
@@ -183,104 +183,128 @@ func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConf
...
 		if err != nil {
-			return errors.Wrap(err, "failed to calculate commit title")
+			logger.Error().Err(err).Msg("Failed to calculate commit title")
+			return
 		}
 		commitMsg.Title = title
 	}
 
-	go func(ctx context.Context) {
-		ticker := time.NewTicker(4 * time.Second)
```
bluekeyes, line 198: I removed the ticker because it doesn't write the first message to the channel until after `duration`, so I think we had an unnecessary 4s delay here.

bulldozer/update.go
```diff
@@ -59,59 +58,43 @@ func ShouldUpdatePR(ctx context.Context, pullCtx pull.Context, updateConfig Upda
...
+func UpdatePR(ctx context.Context, pullCtx pull.Context, client *github.Client, updateConfig UpdateConfig, baseRef string) {
 	logger := zerolog.Ctx(ctx)
 
-	//todo: should the updateConfig struct provide any other details here?
-
-	go func(ctx context.Context, baseRef string) {
-		ticker := time.NewTicker(2 * time.Second)
-		defer ticker.Stop()
-
-		for i := 0; i < MaxPullRequestPollCount; i++ {
```
bluekeyes, line 71: As far as I can tell, this loop and the ticker were dead code: no part of the loop body used `continue` and the final statement in the loop is `return`.

go.mod
```diff
@@ -5,11 +5,11 @@ go 1.13
 require (
 	github.com/c2h5oh/datasize v0.0.0-20171227191756-4eba002a5eae
 	github.com/die-net/lrucache v0.0.0-20181227122439-19a39ef22a11
-	github.com/google/go-github/v30 v30.1.0
+	github.com/google/go-github/v31 v31.0.0
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/palantir/go-baseapp v0.2.0
-	github.com/palantir/go-githubapp v0.2.0
+	github.com/palantir/go-githubapp v0.2.1-0.20200423005854-8924ef0f1e86
```
bluekeyes, line 12: Temporarily using a branch dependency because I'd like to validate async dispatch in a real application before tagging a `go-githubapp` release.

bulldozer/merge.go
```diff
@@ -153,7 +153,7 @@ func (m *PushRestrictionMerger) DeleteHead(ctx context.Context, pullCtx pull.Con
 // MergePR spawns a goroutine that attempts to merge a pull request. It returns
 // an error if an error occurs while preparing for the merge before starting
 // the goroutine.
-func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConfig MergeConfig) error {
+func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConfig MergeConfig) {
```
jmcampanini, line 155: the godoc is inaccurate

server/config.go
```diff
@@ -45,6 +46,11 @@ type CacheConfig struct {
 	MaxSize datasize.ByteSize `yaml:"max_size"`
 }
 
+type WorkerConfig struct {
```
ajlake, line 49: maybe go-githubapp can provide configuration object(s) for the different dispatcher options? Then servers can reuse the objects (like with baseapp.HTTPConfig)

Then callers can do `githubapp.WithSchedulerFromConfig(config.XXX)`.

bulldozer/merge.go
```diff
@@ -183,104 +183,128 @@ func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConf
...
+			logger.Debug().Msgf("Not deleting refs/heads/%s, delete after merge is not enabled", head)
+		}
+	}
+}
 
-			if mergeState.Closed {
-				logger.Debug().Msg("Pull request already closed")
-				return
-			}
+// attempMerge attempts to merge a pull request, logging any errors and
```
ajlake: `attemptMerge`

server/config.go
bluekeyes: Hmm, right now, the `QueueAsyncScheduler` is the only one with options that users might adjust. I don't think we want to make the scheduler type itself configurable, as it can have a pretty big impact on application functionality.

Caching is another area where we current defer to end applications for configuration. I'll think about extracting some of these structs, but I'm not sure it will by worthwhile.

bluekeyes: I think I'm going to leave the config stuct here for now, it feels like it doesn't really fit with the other configuration code currently in `go-githubapp`.

asvoboda (approved):
I'm happy to try this out!

bluekeyes:
// hey policy-bot!

PR #182, opened 2020-04-28
Improve parts of the example server config

bluekeyes:
Add a section for the new worker options and comment out some options
that have reasonable defaults or are uncommon to avoid confusion with
people thinking they are required or setting invalid values.


ajlake:
👍

PR #189, opened 2020-06-11
Add the default configuration file to the Docker image

sideeffffect:
This will enable running the Docker image straight away (you can do all the necessary configuration by defining environment variables).

Without the file, the server refuses to start.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,6 @@ jobs:
       - image: circleci/golang:1.13.4-stretch
     steps:
       - checkout
-      - run: cp -r config docker/
       - setup_remote_docker
       - attach_workspace:
           at: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
       - image: circleci/golang:1.13.4-stretch
     steps:
       - checkout
-      - run: cp config docker/
+      - run: cp -r config docker/
       - setup_remote_docker
       - attach_workspace:
           at: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ jobs:
       - image: circleci/golang:1.13.4-stretch
     steps:
       - checkout
+      - run: cp config docker/
       - setup_remote_docker
       - attach_workspace:
           at: .
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ FROM scratch
 STOPSIGNAL SIGINT
 
 # add the default configuration file
-COPY config/bulldozer.example.yml /secrets/bulldozer.yml
+COPY ../config/bulldozer.example.yml /secrets/bulldozer.yml
 
 # add static files
 COPY ca-certificates.crt /etc/ssl/certs/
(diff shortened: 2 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @sideeffffect! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asvoboda:
Hey @sideeffffect,

Thanks for the PR. Generally we don't want to include the configuration yml inside the container because it will contain secrets (the app private key, webhook secret, etc) that are environment specific and need to be supplied at runtime instead of compile time.

Usually we would suggest running this with a container scheduler (such as nomad, kubernetes, or even plain docker-compose) and mounting the configuration inside when you deploy the application.

sideeffffect:
Running this in k8s is precisely my use case. I just want to spin up a container from `palantirtechnologies/bulldozer` and configure it from environment variables (because, as you say, some of the values are secrets). Sadly this is not possible, because the server refuses to start, when there is no `/secrets/bulldozer.yml` in the image, even though I don't want to change any configuration via `bulldozer.yml`. So there has to be _some_ bulldozer.yml to appease the server executable. (There are no actual secrets in the proposed `/secrets/bulldozer.yml`, so there is no problem having it there.)
People who are not happy with the content of the current `/secrets/bulldozer.yml` can trivially provide their own.

sideeffffect:
Ultimately what I would like to achieve is for the reference Docker image to be runnable right off the bat.
Providing the missing `/secrets/bulldozer.yml` seemed like the simplest way to achieve that. But I've learned about Bulldozer just recently and am not really familiar with it and its code base, so do you see a better way to achieve that @asvoboda ?
Thank you :)

.circleci/config.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - image: circleci/golang:1.13.4-stretch
     steps:
       - checkout
+      - run: cp -r config docker/
```
bluekeyes: Instead of copying this only in the Circle build, you can modify [the script](https://github.com/palantir/bulldozer/blob/develop/godel/config/dist-plugin.yml#L28-L30) that the build tool, Godel, runs before building the image. That way, it will work for local development too.

bluekeyes (changes requested):
Thanks for proposing this! I agree that it makes sense for the default container to have some configuration, but since the environment variable support was added later, we never updated the image. I think the main risk is that the errors from not having variable set (and getting the default values) will be more confusing than the error about missing a configuration file, but we should probably fix that with validation that would benefit all users.

I've suggested an alternate way to copy the file that works outside of CI.

.circleci/config.yml
sideeffffect: :heavy_check_mark:

sideeffffect:
Cool, thanks for the pointer, I've updated the PR.
Another option is that the (now included) `/secrets/bulldozer.yml` wouldn't contain values which are surely wrong, like `integration_id`, `webhook_secret` or `private_key`, they could be commented out (on the other hand, `web_url` or `v3_api_url`, because they can be valid choices for some scenarios).

bluekeyes (approved):

sideeffffect:
Can you guys please release a new Docker image with this change?

bluekeyes:
https://github.com/palantir/bulldozer/releases/tag/v1.9.1

PR #194, opened 2020-07-15
Account for special GitHub behavior with Travis CI

### LINKED ISSUE #190: Problem with Travis CI check
Hello,

I'm not able to resolve issue if Travis CI checks are enabled for github repo. It consists of two checks (`continuous-integration/travis-ci/pr` and `continuous-integration/travis-ci/push`) and even if passed bulldozer thinks that it is not true and produce following message

```xxxx/policy-bot-test#2 is deemed not mergeable because of unfulfilled status checks: [continuous-integration/travis-ci]"```

The strange is that it complains on not complete name of check - looks like a parent group ?

image

bluekeyes:
Travis CI status checks are apparently special-cased in GitHub code for protected branches such that they behave differently from all other checks. To allow Bulldozer to work for users of Travis CI Enterprise, implement a matching special case in Bulldozer.

Fixes #190.

### DIFF
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -46,21 +46,39 @@ func IsPRWhitelisted(ctx context.Context, pullCtx pull.Context, config Signals) 
 	return matches, reason, err
 }
 
-// setDifference returns all elements in set1 that
-// are not in set2.
-func setDifference(set1, set2 []string) []string {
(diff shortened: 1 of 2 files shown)


asvoboda (approved):
This is an impressively silly bug if GH is actually special casing travis under the hood..

PR #186, opened 2020-06-05
Introduce Config V2 using allowlist and blocklist

adamliesko:
Allowlist and blocklist replace whitelist and blacklist.

- Scrub the README of whitelist/blacklist references, fully adopt config V2
in it.

### Why?

I identify with https://www.reddit.com/r/golang/comments/gy9ylr/go_has_removed_all_uses_of_blacklistwhitelist_and/fte1zk0/.

PR #196, opened 2020-07-24
Replace use of whitelist/blacklist with trigger/ignore

asvoboda:
An alternative approach to #186 without creating a new configuration format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ status checks are successful and required reviews are provided.
 
 Additionally, `bulldozer` can:
 
-- Only merge pull requests that match a allowlist condition, like having a
+- Only merge pull requests that match certain conditions, like having a
   specific label or comment
-- Ignore pull requests that match a denylist condition, like having a specific
+- Ignore pull requests that match certain conditions, like having a specific
   label or comment
 - Automatically keep pull request branches up-to-date by merging in the target
   branch
@@ -219,7 +219,7 @@ which are to be expected, and others that may be caused by mis-configuring Bulld
   repository settings
 * Branch protection rules are preventing `bulldozer[bot]` from [pushing to the
   branch][push restrictions]. Github apps can be added to the list of restricted
-  push users, so you can allowlist bulldozer specifically for your repo.
+  push users, so you can allow Bulldozer specifically for your repo.
 
 [push restrictions]: https://help.github.com/articles/about-branch-restrictions/
 [a workaround]: #can-bulldozer-work-with-push-restrictions-on-branches
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ including required status checks and required reviews, are respected. It also
 means that you _must_ enable branch protection to prevent `bulldozer` from
 immediately merging every pull request.
 
-Only pull requests matching the allowlist conditions (or _not_ matching the
-denylist conditions) are considered for merging. `bulldozer` is event-driven,
+Only pull requests matching the trigger conditions (or _not_ matching
+ignore conditions) are considered for merging. `bulldozer` is event-driven,
 which means it will usually merge a pull request within a few seconds of the
 pull request satisfying all preconditions.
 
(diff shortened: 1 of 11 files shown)


bulldozer/config_v1.go
```diff
@@ -34,8 +34,10 @@ const (
 )
 
 type MergeConfig struct {
-	Whitelist Signals `yaml:"whitelist"`
-	Blacklist Signals `yaml:"blacklist"`
+	Allowlist Signals  `yaml:"allowlist"`
+	Denylist  Signals  `yaml:"denylist"`
+	Blacklist *Signals `yaml:"blacklist,omitempty"`
+	Whitelist *Signals `yaml:"whitelist,omitempty"`
```
asvoboda: I would prefer something like.. 
```
	Allowlist Signals  `yaml:"allowlist,whitelist"`
```

bulldozer/config_fetcher.go
```diff
@@ -148,6 +148,20 @@ func (cf *ConfigFetcher) unmarshalConfig(bytes []byte) (*Config, error) {
 		return nil, errors.Wrapf(err, "failed to unmarshal configuration")
 	}
 
+	// Merge old signals configurations if they exist
+	if config.Merge.Blacklist != nil {
+		config.Merge.Denylist = *config.Merge.Blacklist
```
bluekeyes: I think this means that configuration setting both blacklist and denylist will use the values from the old fields. Is that the desired behavior or should the new fields win if both are set?

bulldozer/config_v1.go
bluekeyes: An alternate option might be to only define the new fields and them implement a custom `UnmarshalYAML` function for these types. That would probably do something like:

1. Unmarshal into the full struct with the new fields
2. Try to unmarshal into a subset struct containing only old fields:
   ```go
   var legacy struct {
     Blacklist *Signals `yaml:"blacklist"`
     Whitelist *Signals `yaml:"whitelist"`
   }
   if err := unmarshal(&legacy); err != nil {
     // ...
   }
3. Merge the fields into the main struct if they are set

This all requires that the `unmarshal` function passed to `UnmarshalYAML` is non-strict (i.e. allows unknown fields), but I think that is the case.

bulldozer/evaluate.go
```diff
@@ -24,23 +24,23 @@ import (
 	"github.com/palantir/bulldozer/pull"
 )
 
-// IsPRBlacklisted returns true if the PR is identified as blacklisted,
+// IsPRDenied returns true if the PR is identified as denylisted,
```
bluekeyes:
```suggestion
// IsPRDenied returns true if the PR is identified as denied,
```

bulldozer/update_test.go
```diff
@@ -66,15 +66,15 @@ func generateUpdateTestCase(blacklistable bool, blacklisted bool, whitelistable
 	pullCtx := pulltest.MockPullContext{}
 
 	if blacklistable {
```
bluekeyes: Worth renaming these variables while you're here?

README.md
```diff
@@ -8,9 +8,9 @@ status checks are successful and required reviews are provided.
 
 Additionally, `bulldozer` can:
 
-- Only merge pull requests that match a whitelist condition, like having a
+- Only merge pull requests that match a allowlist condition, like having a
```
bluekeyes:
```suggestion
- Only merge pull requests that match certain conditions, like having a
```

README.md
```diff
@@ -8,9 +8,9 @@ status checks are successful and required reviews are provided.
...
   specific label or comment
-- Ignore pull requests that match a blacklist condition, like having a specific
+- Ignore pull requests that match a denylist condition, like having a specific
```
bluekeyes:
```suggestion
- Ignore pull requests that match certain conditions, like having a specific
```

README.md
```diff
@@ -219,7 +219,7 @@ which are to be expected, and others that may be caused by mis-configuring Bulld
   repository settings
 * Branch protection rules are preventing `bulldozer[bot]` from [pushing to the
   branch][push restrictions]. Github apps can be added to the list of restricted
-  push users, so you can whitelist bulldozer specifically for your repo.
+  push users, so you can allowlist bulldozer specifically for your repo.
```
bluekeyes:
```suggestion
  push users, so you can allow bulldozer specifically for your repo.
```

bulldozer/config_v1.go
```diff
@@ -60,8 +62,10 @@ type SquashOptions struct {
 }
 
 type UpdateConfig struct {
-	Whitelist Signals `yaml:"whitelist"`
-	Blacklist Signals `yaml:"blacklist"`
+	Allowlist Signals  `yaml:"allowlist"`
+	Denylist  Signals  `yaml:"denylist"`
+	Blacklist *Signals `yaml:"blacklist,omitempty"`
+	Whitelist *Signals `yaml:"whitelist,omitempty"`
```
bluekeyes: If you stick with this approach instead of a custom unmarshal function, let's add comments indicating these fields are deprecated and will be removed in a future v2 format.

bluekeyes (commented):
To me, "deny" doesn't feel like quite the right concept for how Bulldozer uses the `denylist` information, but the standard symmetry with "allow" might outweigh that. I think "allow" (or maybe "enable") and "ignore" better match how Bulldozer behaves, but are certainly less standard.

I guess let's check with the rest of the team and then stick with allow/deny unless other people feel strongly (and sorry if we already discussed this and I forgot.)

jamestoyer (approved):

bulldozer/config_fetcher.go
asvoboda: You're right, I'll update this to only accept the legacy values when the new structs aren't present.

asvoboda:
Unfortunately it looks like we use `unmarshalStrict` for the config. I'll go with the existing approach for now.

bluekeyes (approved):

PR #200, opened 2020-08-21
Accept wildcards in branch names

### LINKED ISSUE #199: Wildcard branch selection
Hi,

GitHub Branch protection policies can be setup with wildcards to protect several branches at once, e.g. `release/*`
Have you considered enabling the same for the `merge.trigger.branches` key in the configuration?

I have been running bulldozer 1.9.1 with a patch to enable wildcard behaviour without any issues, but I wonder if there is any appetite for this feature among the maintainers?

-Andrey

andreyzher:
Closes #199 

Enable selection of branches to watch for merges/updates using regex.

Signed-off-by: Andrey Zhereshchin <andrey.zhereshchin@ie.ibm.com>

### DIFF
--- a/bulldozer/signals.go
+++ b/bulldozer/signals.go
@@ -17,6 +17,7 @@ package bulldozer
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/rs/zerolog"
@@ -113,7 +114,7 @@ func (s *Signals) Matches(ctx context.Context, pullCtx pull.Context, tag string)
 		logger.Debug().Msgf("No branches found to match against")
 	}
 	for _, signalBranch := range s.Branches {
-		if targetBranch == signalBranch {
+		if matched, _ := regexp.MatchString(signalBranch, targetBranch); matched {
 			return true, fmt.Sprintf("pull request target is a %s branch: %q", tag, signalBranch), nil
 		}
 	}
--- a/bulldozer/signals.go
+++ b/bulldozer/signals.go
@@ -31,6 +31,7 @@ type Signals struct {
 	Comments          []string `yaml:"comments"`
 	PRBodySubstrings  []string `yaml:"pr_body_substrings"`
 	Branches          []string `yaml:"branches"`
+	BranchPatterns    []string `yaml:"branch_patterns"`
 }
 
 func (s *Signals) Enabled() bool {
@@ -40,6 +41,7 @@ func (s *Signals) Enabled() bool {
 	size += len(s.Comments)
 	size += len(s.PRBodySubstrings)
 	size += len(s.Branches)
+	size += len(s.BranchPatterns)
 	return size > 0
 }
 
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @andreyzher! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bulldozer/signals.go
```diff
@@ -113,7 +114,7 @@ func (s *Signals) Matches(ctx context.Context, pullCtx pull.Context, tag string)
 		logger.Debug().Msgf("No branches found to match against")
 	}
 	for _, signalBranch := range s.Branches {
-		if targetBranch == signalBranch {
+		if matched, _ := regexp.MatchString(signalBranch, targetBranch); matched {
```
bluekeyes: As mentioned on the linked issue, I think we'll want to use a new `BranchPatterns` key for this functionality to avoid surprises with existing configuration. Since `regexp.MatchString` finds any matches, existing names that used to match full branches can now match substrings. We could avoid that by surrounding the pattern with `^` and `$` (if they aren't already present), but that doesn't handle branches that had special characters (`.` would probably be most common.)

bulldozer/signals_test.go
```diff
@@ -31,7 +31,7 @@ func TestSignalsMatches(t *testing.T) {
 		Comments:          []string{"FULL_COMMENT_PLZ_MERGE"},
 		CommentSubstrings: []string{":+1:"},
 		PRBodySubstrings:  []string{"BODY_MERGE_PLZ"},
-		Branches:          []string{"develop"},
+		Branches:          []string{"develop", "test/*"},
```
bluekeyes: This should be `test/.*` since we're using regular expressions, while the GitHub interface uses glob patterns.

bluekeyes (commented):
Thanks for implementing this! I think a separate property will be best to avoid unexpected behavior for existing configuration.

bulldozer/signals.go
andreyzher: Thank you for your feedback. I separated the logic out to a separate loop over `BranchPatterns`, and added a couple of extra tests to make sure there are no unintended partial matches.

bulldozer/signals_test.go
andreyzher: Updated to use regular expression syntax, though it is strange that `regexp.MatchString(a, b)` also accepts glob patterns - the initial version of this PR is what I had been running internally in our team.

(Golang is new to me, so this "dual" behaviour seems odd)

bluekeyes: `regexp.MatchString` does not accept glob patterns, but every glob pattern is a valid regular expression (though with a different meaning.) Because it finds a match at any position (and does not require a complete match), using `test/*` (the string `test` followed by 0 or more `/` characters) matches the expected name but also matches other things:

- `test/my-branch`
- `test//////my-branch`
- `testing`
- `not-a-test`

bluekeyes (approved):
Thanks for the updates, this looks good. Once it merges, you can start using the `snapshot` Docker image if you'd like and then I'll try to make a new release in the next week or two.

PR #213, opened 2020-11-20
Leave a comment when Bulldozer is unable to merge a PR

### LINKED ISSUE #195: Notify user when merge fails
Currently, when bulldozer fails to merge it does so silently behind the scenes.

Instead, we can post a comment to the PR to indicate what happened. The error string included from GH is quite helpful in this case, and we log it already.

```
Merge rejected due to unsatisfied condition: \"You're not authorized to push to this branch. Visit https://help.github.com/enterprise/2.20/user/articles/about-protected-branches/ for more information.\
```

asvoboda:
Resolves #195 

image

### DIFF
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -16,6 +16,7 @@ package bulldozer
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


bluekeyes:
With any change like this we'll have to be extra careful to make sure Bulldozer is reporting this information in a minimal way to avoid overloading PRs with noise. I think the current implementation will lead to annoying amounts of comment spam (unless I'm missing something about how it works.)

`ProcessPullRequest` is called in every event handler and the event handlers mostly aren't doing pre-filtering of events. That's because `ShouldMergePR` is the primary filter function determining if Bulldozer should do any work for this event. Reporting the reason returned by `ShouldMergePR` every time it is not empty (i.e. on every event that doesn't result in a merge) will be excessive. The order in which `ShouldMergePR` checks conditions also doesn't help. For instance, checking for required status last means we'll leave comments on PRs where the GitHub UI is already showing that merging is not possible. A few other scenarios:

1. A repository with Bulldozer enabled in trigger mode will get a comment on every opened PR that doesn't start out with the trigger conditions. This comment will repeat for every comment, review, successful status, push, etc. even if the user never intends to enable the Bulldozer trigger. 
2. A repository with Bulldozer enabled in ignore mode will get a comment from Bulldozer for every event that happens before the build completes. A project that uses multiple jobs in the Circle workflow will get a comment each time a new job finishes saying that the remaining jobs are still pending.

Looking at our logs for the string `deemed not mergeable` over the last 24 hours shows an average of 20 messages (with median of 15 and 95th percentile of 55) _per_ pull request. We definitely don't want every one of those log messages to become a comment.

Last, I don't think this fully solves the original problem in #195, since it doesn't handle the case when merging actually fails, only the case where we decided not to attempt a merge.

My preference would be to focus on the error case first. It's the hardest to figure out without log access and I think it will also be easier to avoid excessive comments. In comparison to the cases covered here, while the comment makes it obvious, a user should be able to figure out what's happening by referencing the in-repo configuration file and the existing GitHub UI elements.

Maybe we could also look at alternate notification methods like statuses if managing the comments proves too difficult.

bluekeyes (changes requested):
While the specifics of this code look fine to me, I have concerns about the overall approach (detailed in my previous comment.)

asvoboda:
Great point(s). I'll close this out and open a new targeted PR for just handling the error condition.

On overloading PRs: Ideally we're only adding a single comment. In this current iteration, we add one comment per event, which in practice is probably too many (as pointed out, one should be sufficient). As part of the next PR I'll look into only leaving a comment iff the most recent comment is not by the bot, which should help for reducing noise.

PR #217, opened 2021-01-26
Add option to prevent merges without required checks

### LINKED ISSUE #122: Do not merge PRs with no required checks by default
Currently, Bulldozer will wait for any required status checks defined in GitHub or in the configuration before attempting a merge, but it does not require that at least one of these checks exists. This means that Bulldozer will immediately merge pull requests on repositories that don't define any required checks, which can get exciting if branch protection is accidentally removed or modified.

By default, Bulldozer should refuse to merge pull requests in any repository that does not specify at least one required check via either branch protection or the configuration. Users who wish to keep the current behavior can set a new property, `allow_merge_with_no_checks`. This will probably be a per-repository configuration option rather than a server option.

derekjobst:
This PR includes changes which seek to resolve #122, where PRs without any required checks are merged automatically, often due to mis-configuration.

A new option is added to the v1 configuration format to allow/deny merges with no required status checks. The option is nested within the `merge` object:
```yaml
version: 1

merge:
  allow_merge_with_no_checks: false
```

The v0 configuration format is not updated to support this feature. Repositories using the v0 configuration will be treated as if `allow_merge_with_no_checks` is set to the default value `false`.

### DIFF
--- a/bulldozer/config_fetcher.go
+++ b/bulldozer/config_fetcher.go
@@ -189,8 +189,9 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				Trigger: Signals{
 					Labels: []string{"merge when ready", "merge-when-ready", "merge_when_ready"},
 				},
-				DeleteAfterMerge: configv0.DeleteAfterMerge,
-				Method:           configv0.Strategy,
+				DeleteAfterMerge:       configv0.DeleteAfterMerge,
+				AllowMergeWithNoChecks: configv0.AllowMergeWithNoChecks,
+				Method:                 configv0.Strategy,
 			},
 		}
 		if config.Merge.Method == SquashAndMerge {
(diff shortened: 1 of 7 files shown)


bulldozer/config_v0.go
```diff
@@ -23,9 +23,10 @@ const (
 )
 
 type ConfigV0 struct {
-	Mode             ModeV0      `yaml:"mode"`
-	Strategy         MergeMethod `yaml:"strategy"`
-	DeleteAfterMerge bool        `yaml:"deleteAfterMerge"`
+	Mode                   ModeV0      `yaml:"mode"`
+	Strategy               MergeMethod `yaml:"strategy"`
+	DeleteAfterMerge       bool        `yaml:"deleteAfterMerge"`
+	AllowMergeWithNoChecks bool        `yaml:"allow_merge_with_no_checks"`
```
bluekeyes: The v0 configuration is deprecated and shouldn't get any new features. People who wish to configure this option should move to the supported v1 configuration.

bluekeyes (commented):
Aside from the v0 config, this looks good to me

bulldozer/config_fetcher.go
```diff
@@ -189,8 +189,9 @@ func (cf *ConfigFetcher) unmarshalConfigV0(bytes []byte) (*Config, error) {
 				Trigger: Signals{
 					Labels: []string{"merge when ready", "merge-when-ready", "merge_when_ready"},
 				},
-				DeleteAfterMerge: configv0.DeleteAfterMerge,
-				Method:           configv0.Strategy,
+				DeleteAfterMerge:       configv0.DeleteAfterMerge,
+				AllowMergeWithNoChecks: false,
```
bluekeyes, line 193: I think it's fine to set these explicitly, but it's not required since the zero value of the field is the desired default value.

bluekeyes (approved):

bulldozer/config_fetcher.go
derekjobst: 👍🏻  I was on the fence about that, but figured it might be better to be explicit about the setting.

PR #229, opened 2021-03-19
Allow server env var prefix to be overridden

f1sherman:
As suggested [here](https://github.com/palantir/bulldozer/pull/220#issuecomment-781764581), use a `BULLDOZER_ENV_PREFIX` environment variable to override the prefix for server configuration values. This makes it possible to run the Docker container on Heroku, where it is expected that the container will listen on the port specified in the `PORT` environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,10 @@ We provide both a Docker container and a binary distribution of the server:
 
 A sample configuration file is provided at `config/bulldozer.example.yml`.
 Certain values may also be set by environment variables; these are noted in the
-comments in the sample configuration file.
+comments in the sample configuration file. By default, the environment
+variables for server values are prefixed with `BULLDOZER_` (e.g.
+`BULLDOZER_PORT`). This prefix can be overridden by setting the
+`BULLDOZER_ENV_PREFIX` environment variable.
 
 ### GitHub App Configuration
 
--- a/server/config.go
+++ b/server/config.go
@@ -27,6 +27,10 @@ import (
 	"github.com/palantir/bulldozer/bulldozer"
 )
 
+const (
+	DefaultEnvPrefix = "BULLDOZER_"
+)
+
 type Config struct {
 	Server  baseapp.HTTPConfig `yaml:"server"`
 	Github  githubapp.Config   `yaml:"github"`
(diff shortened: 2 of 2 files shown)


f1sherman:
@bluekeyes I'm wondering if there is any chance that you're still open to this change as we discussed previously? Despite confirming with Heroku support that I'd be able to set the `BULLDOZER_PORT` based on `PORT` in the `.profile`, I haven't been able to get that working and am now wondering if something was lost in communication. I'm working with Heroku support on this to confirm, but having this change would help me get unblocked. Sorry for the bother and thanks again for all of your help!

f1sherman:
I was able to get this working in Heroku by using the binary and building my own Docker container, so feel free to close this if you don't feel it would be a worthwhile change!

server/config.go
```diff
@@ -67,7 +67,13 @@ func ParseConfig(bytes []byte) (*Config, error) {
...
 	c.Github.SetValuesFromEnv("")
-	c.Server.SetValuesFromEnv("BULLDOZER_")
+
+	serverPrefix := "BULLDOZER_"
+	if v, ok := os.LookupEnv("BULLDOZER_ENV_PREFIX"); ok {
+		serverPrefix = v
+	}
+	c.Server.SetValuesFromEnv(serverPrefix)
+
 	if v, ok := os.LookupEnv("BULLDOZER_PUSH_RESTRICTION_USER_TOKEN"); ok {
```
bluekeyes:
```suggestion
	if v, ok := os.LookupEnv(envPrefix + "PUSH_RESTRICTION_USER_TOKEN"); ok {
```

I think we should use the custom prefix for all prefixed variables.

server/config.go
```diff
@@ -67,7 +67,13 @@ func ParseConfig(bytes []byte) (*Config, error) {
 	}
 
 	c.Github.SetValuesFromEnv("")
-	c.Server.SetValuesFromEnv("BULLDOZER_")
+
+	serverPrefix := "BULLDOZER_"
```
bluekeyes: As a style note, I'd like to call this `envPrefix` and make the default value a constant at the top of this file:

```go
const (
    DefaultEnvPrefix = "BULLDOZER_"
)
```

README.md
```diff
@@ -288,7 +288,10 @@ We provide both a Docker container and a binary distribution of the server:
 
 A sample configuration file is provided at `config/bulldozer.example.yml`.
 Certain values may also be set by environment variables; these are noted in the
-comments in the sample configuration file.
+comments in the sample configuration file. By default, the environment
+variables for server values are prefixed with `BULLDOZER_` (e.g.
+`BULLDOZER_PORT`). This prefix can be overridden by setting the
+`BULLDOZER_ENV_PREFIX` environment variable.
```
bluekeyes, line 294: Thanks for updating the documentation!

bluekeyes (commented):
Sorry for the delay reviewing this. While it is a bit of a strange feature, I don't think it hurts long-term maintainability so I'm not opposed to adding it if it makes the standard Docker image more usable.

f1sherman:
Thanks for the feedback @bluekeyes! I believe I've made all of the updates you've requested. As a bonus this will make it easier for me to test #235 as well!

bluekeyes (approved):

f1sherman:
@bluekeyes just an update that I've been testing this today and it's working well!

Issue #145, opened 2019-08-10
update is not triggered on label change, comment etc


misha-ridge:
PR update is only triggered on "push", so in the following sequence of events PR will never be updated/merged:

- Alice creates a branch `alice-feature` off `master`
- Bob publishes a `bob-feature` PR, and it gets merged
- Alice publishes the `alice-feature` PR without rebasing/merging `master` into it
- Status checks for `alice-feature` PR pass
- ...?

bluekeyes:
Thanks for reporting this. Bulldozer's update behavior is a known limitation (see #111), but given the potential for a flood of CI builds and the other confusing behavior (#128, #95, #66), I'm reluctant to expand it to too many additional triggers. 

Right now, I think it would be reasonable to consider updates when:

- A PR is first opened (this handles the situation in original ticket, provided the PR is opened with a label or comment in the body)
- When a (matching) label or comment is added (this provides a manual trigger)

I'm less sure that it is useful to consider updates in response to reviews or status checks, as the current PR proposes.

misha-ridge:
> I'm less sure that it is useful to consider updates in response to reviews or status checks, as the current PR proposes.

Then the documentation should be updated to explicitly say so: "updates are performed only when a base branch is updated". I find it very confusing that the behaviour of update changes based on actual sequencing of events "label, then push to master" vs. "push to master then label", but probably not everyone shares this point of view.

Regarding tickets:

>  #95 

Is another feature. I personally don't see much logic in it (why does one care about the _number_ of updates performed if the branch ends up being merged at the end?), but probably in some workflows is is useful.

> #66 

Is just a bug.

> #128 

Well, this is a fine kettle of fish. Merging N PRs that all have to be up-to-date to base branch is going to be computationally expensive (O(N^2)) if we run CI on every update. However, it is already so in current implementation, so I don't see how adding missing triggers is going to change it.

brendan-donegan:
My team is finding this behaviour deeply confusing, it would be really great if adding a tag triggered Bulldozer to check if the PR is out of date or mergeable

jlewi:
> I'm less sure that it is useful to consider updates in response to reviews or status checks, as the current PR proposes.

If you don't update in response to status checks does this create a problem for automatic processes?

We have automated processes open PRs up against a repo with automerge label attached. We'd like those PRs to be auto-merged if the tests pass.

[...]

PR #235, opened 2021-04-06
Update PR when labeled/opened

f1sherman:
This will conditionally update a PR whenever the PR is labeled or opened, which will partially fulfill [this
suggestion](https://github.com/palantir/bulldozer/issues/145#issuecomment-520662592) for https://github.com/palantir/bulldozer/issues/145.

It doesn't seem possible to determine which label was added via the Github webhook payload, so this will update the PR regardless of whether the label added is what is configured in `bulldozer.yml`.

I haven't had a chance to test this yet (will update when I have), so I'm just looking for early feedback at this point. Thanks!

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -33,12 +33,33 @@ type Base struct {
 	PushRestrictionUserToken string
 }
 
-func (b *Base) ProcessPullRequest(ctx context.Context, pullCtx pull.Context, client *github.Client, pr *github.PullRequest) error {
+func (b *Base) FetchConfig(ctx context.Context, client *github.Client, pr *github.PullRequest) (*bulldozer.Config, error) {
 	logger := zerolog.Ctx(ctx)
 
 	bulldozerConfig, err := b.ConfigForPR(ctx, client, pr)
 	if err != nil {
-		return errors.Wrap(err, "failed to fetch configuration")
+		return nil, errors.Wrap(err, "failed to fetch configuration")
+	}
+
+	switch {
+	case bulldozerConfig.Missing():
+		logger.Debug().Msgf("No configuration found for %s", bulldozerConfig)
+		return nil, nil
+	case bulldozerConfig.Invalid():
+		logger.Warn().Msgf("Configuration is invalid for %s", bulldozerConfig)
+		return nil, nil
+	}
+
+	logger.Debug().Msgf("Found valid configuration for %s", bulldozerConfig)
+	return bulldozerConfig.Config, nil
+}
+
+func (b *Base) ProcessPullRequest(ctx context.Context, pullCtx pull.Context, client *github.Client, config *bulldozer.Config, pr *github.PullRequest) error {
+	logger := zerolog.Ctx(ctx)
+
+	if config == nil {
+		logger.Debug().Msg("ProcessPullRequest: returning immediately due to nil config")
+		return nil
 	}
 
 	merger := bulldozer.NewGitHubMerger(client)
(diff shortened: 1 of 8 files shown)


server/handler/pull_request.go
```diff
@@ -66,6 +67,21 @@ func (h *PullRequest) Handle(ctx context.Context, eventType, deliveryID string,
 		logger.Error().Err(errors.WithStack(err)).Msg("Error processing pull request")
 	}
 
+	if event.GetAction() == "labeled" || event.GetAction() == "opened" {
+		logger.Debug().Msgf("received pull_request %s event", event.GetAction())
```
bluekeyes: This logging is probably more appropriate higher up in the function where it will apply for all possible actions, rather than only `labeled` and `opened` actions.

server/handler/pull_request.go
```diff
@@ -66,6 +67,21 @@ func (h *PullRequest) Handle(ctx context.Context, eventType, deliveryID string,
...
+
+		base, _ := pullCtx.Branches()
+		ref, _, err := client.Git.GetRef(ctx, pullCtx.Owner(), pullCtx.Repo(), fmt.Sprintf("refs/heads/%s", base))
+		if err != nil {
+			return errors.Wrap(err, "could not get git reference of PR base branch")
+		}
+
+		baseRef := ref.GetObject().GetSHA()
+		if err := h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, pr, baseRef); err != nil {
```
bluekeyes: The `baseRef` passed to `UpdatePullRequest` is a ref name rather than a SHA. You should be able to use the `fmt.Sprintf("refs/heads/%s", base)` directly, without resolving it.

bluekeyes (commented):
Thanks for adding this. Based on the Go documentation for [`PullRequestEvent`](https://pkg.go.dev/github.com/google/go-github/v32/github#PullRequestEvent), there is a `Label` field that is populated for `labeled` actions. That said, the current structure of the code makes it difficult to use this value for anything, so I think we can try the simple approach of doing this evaluation on every `labeled` action for now.

The bigger refactor (which you might not want to take on) would be to change `Base#ProcessPR` and `Base#UpdatePR` so that we don't duplicate configuration loading. As is, we'll look for and load configuration twice, although the requests the second time should hit the HTTP cache.

server/handler/pull_request.go
f1sherman: Great idea, will move this further up

server/handler/pull_request.go
f1sherman: Nice, that will really simplify things

f1sherman:
@bluekeyes I believe I've addressed all of your feedback. I also took a shot at that refactor as a separate commit, let me know if you want me to change that or remove it. Thanks!

server/handler/base.go
```diff
@@ -33,14 +33,28 @@ type Base struct {
...
 
 	bulldozerConfig, err := b.ConfigForPR(ctx, client, pr)
 	if err != nil {
-		return errors.Wrap(err, "failed to fetch configuration")
+		return nil, errors.Wrap(err, "failed to fetch configuration")
 	}
 
+	switch {
+	case bulldozerConfig.Missing():
+		return nil, errors.Errorf("No configuration found for %s", bulldozerConfig)
```
bluekeyes: One consequence of this conversion is that these conditions that were previously only logged are now returned and propagated as errors. I think I'd like to keep the old behavior by logging the same messages here and then returning `nil, nil`.

This means that either callers will have to check for `nil` or `UpdatePR` and `ProcessPR` should return immediately if passed a `nil` configuration; I'd probably have the functions return immediately.

server/handler/base.go
```diff
@@ -33,14 +33,28 @@ type Base struct {
...
+	case bulldozerConfig.Invalid():
+		return nil, errors.Errorf("Configuration is invalid for %s", bulldozerConfig)
+	}
+
+	logger.Debug().Msgf("Found valid configuration for %s", bulldozerConfig)
+	config := bulldozerConfig.Config
+
+	return config, nil
```
bluekeyes: style:
```suggestion
	return bulldozerConfig.Config, nil
```

server/handler/check_run.go
```diff
@@ -73,7 +73,11 @@ func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, pay
 		pullCtx := pull.NewGithubContext(client, fullPR)
 
 		logger := logger.With().Int(githubapp.LogKeyPRNum, pr.GetNumber()).Logger()
-		if err := h.ProcessPullRequest(logger.WithContext(ctx), pullCtx, client, fullPR); err != nil {
+		config, err := h.FetchConfig(ctx, client, pr)
+		if err != nil {
+			return errors.Wrap(err, "failed to fetch configuration")
```
bluekeyes: Because `FetchConfig` already wraps errors with a similar message, wrapping them again doesn't add useful context.
```suggestion
			return err
```

server/handler/issue_comment.go
```diff
@@ -57,7 +57,11 @@ func (h *IssueComment) Handle(ctx context.Context, eventType, deliveryID string,
 	}
 	pullCtx := pull.NewGithubContext(client, pr)
 
-	if err := h.ProcessPullRequest(ctx, pullCtx, client, pr); err != nil {
+	config, err := h.FetchConfig(ctx, client, pr)
+	if err != nil {
+		return errors.Wrap(err, "failed to fetch configuration")
```
bluekeyes:
```suggestion
		return err
```

server/handler/pull_request.go
```diff
@@ -62,7 +64,19 @@ func (h *PullRequest) Handle(ctx context.Context, eventType, deliveryID string,
 	}
 	pullCtx := pull.NewGithubContext(client, pr)
 
-	if err := h.ProcessPullRequest(ctx, pullCtx, client, pr); err != nil {
+	config, err := h.FetchConfig(ctx, client, pr)
+	if err != nil {
+		return errors.Wrap(err, "failed to fetch configuration")
```
bluekeyes:
```suggestion
		return err
```

server/handler/status.go
```diff
@@ -68,7 +68,11 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
 	for _, pr := range prs {
 		pullCtx := pull.NewGithubContext(client, pr)
 		logger := logger.With().Int(githubapp.LogKeyPRNum, pr.GetNumber()).Logger()
-		if err := h.ProcessPullRequest(logger.WithContext(ctx), pullCtx, client, pr); err != nil {
+		config, err := h.FetchConfig(ctx, client, pr)
+		if err != nil {
+			return errors.Wrap(err, "failed to fetch configuration")
```
bluekeyes:
```suggestion
			return err
```

bluekeyes (commented):
Thanks for taking on the config refactor too. All of my comments relate to that; the PR handler part of this change looks good. If you'd like to fix up the refactor part, that would be great, otherwise, feel free to remove that commit and I'll clean up some of the duplication in a separate change.

f1sherman:
@bluekeyes ok I went ahead and made those updates. Thanks so much for patiently working with me through those. Is this something you'd like me to test before merging or would it make more sense to merge first then test? Merging first makes testing a little easier for me but I'm ok with either way.

bluekeyes (approved):

bluekeyes:
This seems relatively safe in terms of changes to existing functionality, so I think we can merge first and then test.

f1sherman:
@bluekeyes just an update that I've been testing this today and it's working well!

f1sherman:
@bluekeyes after some more testing I noticed a bug with this change (or at least I consider it a bug): when a webhook triggers both an update and a merge, Bulldozer continues to use the stale statuses from before the update to evaluate whether a merge can happen. I'll work on putting together a PR to fix this.

f1sherman:
@bluekeyes I put together a proposed fix for this issue [here](https://github.com/palantir/bulldozer/pull/242). Feedback welcome (as always)!

PR #243, opened 2021-04-23
Remove delete_after_merge functionality

### LINKED ISSUE #212: Remove delete_after_merge option
GH now supports automatically deleting branches (as of 2.22.4), so Bulldozer no longer needs to support the delete_after_merge option.

We can probably remove the supporting code and leave the option around, if we don't want to break any config changes:
https://github.com/palantir/bulldozer/blob/a49528932f30a3d1666ccc35c2bd28f3959422b4/bulldozer/merge.go#L215-L216

anishagg17:
Fixes: #212 

removed code for attemptDelete and DeleteAfterMerge functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,6 @@ merge:
   required_statuses:
     - "ci/circleci: ete-tests"
 
-  # If true, bulldozer will delete branches after their pull requests merge.
-  delete_after_merge: true
-
   # If true, bulldozer will merge pull requests with no required checks. This
   # helps to protect against merging branches which inadvertently do not have
   # required status checks.
--- a/bulldozer/config_fetcher_test.go
+++ b/bulldozer/config_fetcher_test.go
@@ -39,7 +39,6 @@ merge:
   options:
     squash:
       body: summarize_commits
-  delete_after_merge: true
 
 update:
   trigger:
@@ -77,7 +76,6 @@ merge:
   options:
     squash:
       body: summarize_commits
-  delete_after_merge: true
 
 update:
   whitelist:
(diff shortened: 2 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @anishagg17! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asvoboda:
Hey @anishagg17, thanks for the PR. 

Do you mind also removing the examples as well? https://github.com/palantir/bulldozer/blob/0bfd9486b8d72b5c9dbdf04d91f09bd1783a48c8/config/examples/standard.bulldozer.v1.yml#L14
 https://github.com/palantir/bulldozer/blob/19a4b0800b00bab36aaf30366800f1016983efb5/README.md#bulldozeryml-specification

anishagg17:
sure, @asvoboda I have removed it from some tests too and even from the `Readme`

anishagg17:
Oh, I think that after changing the `test config` I need to make changes to the expected output also. Am I correct?

asvoboda:
If you run `./godelw verify` locally, you can see the same output from the test.

It looks like the `head` var L158 in merge.go is now unused, and the compiler is complaining.

anishagg17:
> It looks like the head var L158 in merge.go is now unused, and the compiler is complaining.

Yeah, replaced it with an `_`.

anishagg17:
@asvoboda can we merge this one?

PR #250, opened 2021-06-04
Fix processing for check run events

bluekeyes:
In 1.13.0, an unrelated refactor used an incomplete pull request object
to fetch configuration, resulting in requests that are missing the
organization name. Because the config isn't loaded, all further
processing for check run events is skipped.

Use the complete PR object and also re-order things to make it less
tempting to use the plain PR object once the full one is available.

### DIFF
--- a/server/handler/check_run.go
+++ b/server/handler/check_run.go
@@ -61,22 +61,23 @@ func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, pay
 	}
 
 	for _, pr := range prs {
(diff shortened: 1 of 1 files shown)


bluekeyes:
@jlewi per your comments on #145, if your CI system uses the Checks API, I think it's likely you are hitting this bug. If the last event on a PR is a `check_run` event, Bulldozer fails with an error that it could not find valid configuration and does no further processing. This is a regression in 1.13.0.

That can also explain the randomness of the failures - if some other event type arrives after the `check_run` event, everything works as expected.

jmcampanini (approved):

jlewi:
@bluekeyes Thank you very much.

PR #256, opened 2021-07-27
Allow ignoring Draft PRs from updates

### LINKED ISSUE #191: Add condition for draft PRs
It would be useful to white/blacklist actions based on draft PR status. For example, to not auto-update PRs if they are in draft status as they clearly require more work.

wrslatz:
Allow ignoring Draft PRs from automated Bulldozer update merge commits. Closes #191.

Features:
- Continue updating Draft PRs by default (current functionality)
- New `ignore_draft_pr` config under `update` block to ignore Draft PRs from updating
- Allow updating Draft PRs if `ignore_draft_pr` is enabled *but* the PR matches a configured `trigger` condition

I built and manually tested this in our environment (GitHub Enterprise 2.22) and it worked as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,10 @@ update:
   # bulldozer. It accepts the same keys as the ignore in the "merge" block.
   ignore:
     labels: ["Do Not Update"]
+
+  # If true, bulldozer will ignore updating draft pull requests, unless they
+  # explicitly match a configured trigger condition.
+  ignore_draft_pr: false
 ```
 
 ## FAQ
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ update:
 
   # If true, bulldozer will ignore updating draft pull requests, unless they
   # explicitly match a configured trigger condition.
-  ignore_draft_pr: false
+  ignore_drafts: false
 ```
 
 ## FAQ
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -68,6 +68,8 @@ type UpdateConfig struct {
 	Trigger Signals `yaml:"trigger"`
 	Ignore  Signals `yaml:"ignore"`
 
+	IgnoreDraftPR bool `yaml:"ignore_draft_pr"`
+
 	// Blacklist and Whitelist are legacy options that will be disabled in a future v2 format
 	Blacklist Signals `yaml:"blacklist"`
 	Whitelist Signals `yaml:"whitelist"`
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -68,7 +68,7 @@ type UpdateConfig struct {
 	Trigger Signals `yaml:"trigger"`
 	Ignore  Signals `yaml:"ignore"`
 
-	IgnoreDraftPR bool `yaml:"ignore_draft_pr"`
+	IgnoreDrafts bool `yaml:"ignore_drafts"`
 
 	// Blacklist and Whitelist are legacy options that will be disabled in a future v2 format
 	Blacklist Signals `yaml:"blacklist"`
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -68,7 +68,7 @@ type UpdateConfig struct {
 	Trigger Signals `yaml:"trigger"`
 	Ignore  Signals `yaml:"ignore"`
 
-	IgnoreDrafts bool `yaml:"ignore_drafts"`
+	IgnoreDrafts *bool `yaml:"ignore_drafts"`
 
 	// Blacklist and Whitelist are legacy options that will be disabled in a future v2 format
 	Blacklist Signals `yaml:"blacklist"`
--- a/pull/github_context.go
+++ b/pull/github_context.go
@@ -290,7 +290,7 @@ func (ghc *GithubContext) IsTargeted(ctx context.Context) (bool, error) {
 }
 
 func (ghc *GithubContext) IsDraft(ctx context.Context) bool {
-	return *ghc.pr.Draft
+	return ghc.pr.GetDraft()
 }
 
 // type assertion
--- a/pull/github_context.go
+++ b/pull/github_context.go
@@ -289,5 +289,9 @@ func (ghc *GithubContext) IsTargeted(ctx context.Context) (bool, error) {
 	return len(prs) > 0, nil
 }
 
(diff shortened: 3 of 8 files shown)


wrslatz:
I considered making `ignore_draft_pr` part of the `ignore` `Signals` under only the `update` config, but that seemed like a more involved change with less value (makes `ignore` `Signals` more confusing depending on the context)

pull/github_context.go
```diff
@@ -289,5 +289,9 @@ func (ghc *GithubContext) IsTargeted(ctx context.Context) (bool, error) {
 	return len(prs) > 0, nil
 }
 
+func (ghc *GithubContext) IsDraft(ctx context.Context) bool {
+	return *ghc.pr.Draft
```
bluekeyes:
```suggestion
	return ghc.pr.GetDraft()
```

README.md
```diff
@@ -165,6 +165,10 @@ update:
   # bulldozer. It accepts the same keys as the ignore in the "merge" block.
   ignore:
     labels: ["Do Not Update"]
+
+  # If true, bulldozer will ignore updating draft pull requests, unless they
+  # explicitly match a configured trigger condition.
+  ignore_draft_pr: false
```
bluekeyes: Since Bulldozer only operates on PRs, I suggest changing this field to `ignore_drafts`.

bulldozer/config_v1.go
```diff
@@ -68,6 +68,8 @@ type UpdateConfig struct {
 	Trigger Signals `yaml:"trigger"`
 	Ignore  Signals `yaml:"ignore"`
 
+	IgnoreDraftPR bool `yaml:"ignore_draft_pr"`
```
bluekeyes: This may need to be a `*bool` depending on how we decided to handle the situation where this is the only set field. See my top-level comment for context.

bluekeyes (commented):
Thanks for implementing this!

> I considered making `ignore_draft_pr` part of the ignore `Signals` under only the `update` config...

I also thought this made sense at first, but the fact that `ignore_draft_pr` can be overridden by `trigger` options (unlike the regular `ignore` block, which always wins) is another argument for having it as separate property. Ideally, it would all be in one place, but I think that's an issue for a future v2 configuration format.

We may need to update the condition on L29 in `bulldozer/update.go` to account for this new field. If `ignore_draft_pr` is the only property set in the `update` block, does that mean that we update all non-draft PRs? Or should the user also have to specify at least one condition in the normal `ignore` block?

wrslatz:
> We may need to update the condition on L29 in `bulldozer/update.go` to account for this new field. If `ignore_draft_pr` is the only property set in the `update` block, does that mean that we update all non-draft PRs? Or should the user also have to specify at least one condition in the normal `ignore` block?

I think it'd be preferable to support ignoring draft PRs with no other conditions set. I'll take a swing at making that change as well as the other suggestions.

bulldozer/config_v1.go
wrslatz: My Go is a bit rusty, could you walk me through why `*bool` would be needed here if it's the only value set? Also how do you think we should handle this case in `bulldozer/update.go`? I'm trying to think through the cleanest implementation path I can

wrslatz: Would it be to differentiate between a `nil` value for the config option vs implicit `false`?

bluekeyes: Yeah, this only needs to be a pointer if there's a reason to differentiate between "not set" and "set to false". With a simple `bool`, there's no way to distinguish between the zero value `false` and an explicit `false` in the config.

The line in `update.go` I mentioned might become something like this:

```go
if !updateConfig.Ignore.Enabled() && !updateConfig.Trigger.Enabled() && udpateConfig.IgnoreDrafts == nil {
	return false, nil
}
```

Then the condition later on would have to check for `nil` first:

```go
if updateConfig.IgnoreDrafts != nil && *updateConfig.IgnoreDrafts && pullCtx.IsDraft(ctx) {
  // ...
}
```

wrslatz:
> I think it'd be preferable to support ignoring draft PRs with no other conditions set. I'll take a swing at making that change as well as the other suggestions.

I read through the `README` description of `update` again and now I'm not so sure
```
# "update" defines how and when to update pull request branches. Unlike with
# merges, if this section is missing, bulldozer will not update any pull requests.
```

It'd be simpler to implement if `IgnoreDrafts` could be kept a `bool`, but describing how `update` config works becomes more difficult in that case. Should having `ignore_drafts: false` as the only config under `update` enable updates for all PRs?

bulldozer/config_v1.go
wrslatz: Cool, that's the exact path I started implementing. Getting the tests right in `bulldozer/update_test.go` is a bit tricky though, as making `ignoreDrafts    *bool` in the `testMatrix` is giving some trouble setting the pointer reference to `true` or `false` boolean cleanly.

wrslatz: I could add a `isIgnoreDraftsNil bool` and keep `ignoreDrafts` as `bool` but that doesn't seem very clean

bluekeyes: Adding a helper function like this is a common way to deal with "literal pointers" in tests:

```go
func boolp(b bool) *bool {
  return &b
}

// in a test
{ true, true, false, boolp(true) }
```

You could also define values and then reference them:

```go
valTrue, valFalse := true, false

// in a test
{ true, true, false, &valTrue }
```

wrslatz: I think I'll go with the former, as the latter all pointing to the same value increases the likelihood of test state being manipulated across tests

bluekeyes:
> Should having `ignore_drafts: false` as the only config under `update` enable updates for all PRs?

Yes, I think that's most consistent with current behavior. If you're unsure of how to deal with the `*bool` in any part of the code, I'm happy to offer some ~pointers~ advice.

bulldozer/config_v1.go
wrslatz: Pushed my attempt at this change, thanks for your ~pointers~ advice

bluekeyes (approved):
Looks good! Thank you for making all the updates from the review.

I'm not sure yet when the next release will be, but you can use the `snapshot` Docker tag if you'd like to start using this before then.

wrslatz:
Thanks for the quick review and the help!

PR #258, opened 2021-08-13
Update go-github to v38, githubapp to v0.9.0

IainSteers:


bulldozer/config_fetcher.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/google/go-github/v32/github"
+	"github.com/google/go-github/v38/github"
```
asvoboda, line 22: In https://github.com/palantir/policy-bot/pull/261 we discovered some weird compatibility with older versions of GHE after bumping to v33 due to removed preview headers. It might be worthwhile calling out that older versions (pre 3.0) of GHE _may_ not work with Bulldozer.

asvoboda (approved):

bulldozer/config_fetcher.go
IainSteers: I think we're safe here as those compat issues were down to the preview headers for Checks API, which Bulldozer doesn't use.

bluekeyes: Bulldozer reads from the Checks API to make sure checks are included for when checking required statuses against the actual statuses: https://github.com/palantir/bulldozer/blob/develop/pull/github_context.go#L238

PR #270, opened 2021-12-04
Add conditional merge configuration

### LINKED ISSUE #266: [FR] Make merge method configurable: merge or squash depending on conditions
Bulldozer currently takes a single 'method', which applies to all PRs regardless of the state of the source branch. 

It would be nice for Bulldozer to do a regular merge when certain conditions are satisfied, and a squash merge when these conditions are not, as specified in the `.palantir/bulldozer.yml` file. These conditions can include the number of commits on the source branch, and the presence/absence of a certain label. 

The configuration could look something like this
```
merge:
  method: squash
  conditional-methods:
    merge:
      labels: ["auto-push", "generated"]
      max-commits: 1
```

This is useful for command line tools that will generate a chain of PRs with a single commit each.

derekjobst:
Closes #266

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ merge:
   # will trigger bulldozer.
   # - This will override any branch_method logic if one of the methods is
   # triggered
+  # - If no trigger criteria is provided the method is ignored
   merge_method:
     # "method" defines the merge method. The available options are "merge",
     # "rebase", "squash", and "ff-only".
--- a/README.md
+++ b/README.md
@@ -139,7 +139,8 @@ merge:
 
   # Allows the merge method that is used when auto-merging a PR to be different
   # based on trigger criteria. The first method where ALL triggers match will
-  # be used. Otherwise, the method specified previously in merge/method will be # used.
+  # be used. Otherwise, the method specified previously in "merge.method" will 
+  # be used.
   # - ALL trigger criteria must match, unlike merge/trigger where ANY match 
   # will trigger bulldozer.
   # - This will override any branch_method logic if one of the methods is
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ merge:
       # If "body" is "pull_request_body", then the commit message will be the
       # part of the pull request body surrounded by "message_delimiter"
       # strings. This is disabled (empty string) by default.
-    message_delimiter: ==COMMIT_MSG==
+      message_delimiter: ==COMMIT_MSG==
 
   # "required_statuses" is a list of additional status contexts that must pass
   # before bulldozer can merge a pull request. This is useful if you want to
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ merge:
       # If "body" is "pull_request_body", then the commit message will be the
       # part of the pull request body surrounded by "message_delimiter"
       # strings. This is disabled (empty string) by default.
-    message_delimiter:bulldozer/config_v1.go ==COMMIT_MSG==
+    message_delimiter: ==COMMIT_MSG==
 
   # "required_statuses" is a list of additional status contexts that must pass
   # before bulldozer can merge a pull request. This is useful if you want to
--- a/README.md
+++ b/README.md
@@ -126,13 +126,33 @@ merge:
   # "rebase", "squash", and "ff-only".
   method: squash
 
-  # Allows the merge method that is used when auto-merging a PR to be different based on the
+  ##### branch_method has been DEPRECATED in favor of merge_method #####
+  # 
+  # Allows the merge method that is used when auto-merging a PR to be different
   # target branch. The keys of the hash are the target branch name, and the values are the merge method that
   # will be used for PRs targeting that branch. The valid values are the same as for the "method" key.
   # Note: If the target branch does not match any of the specified keys, the "method" key is used instead.
   branch_method:
     develop: squash
     master: merge
+  ##### branch_method has been DEPRECATED in favor of merge_method #####
+
+  # Allows the merge method that is used when auto-merging a PR to be different
+  # based on trigger criteria. The first method where ALL triggers match will
+  # be used. Otherwise, the method specified previously in merge/method will be # used.
+  #
+  # Note: ALL trigger criteria must match, unlike merge/trigger where ANY match 
+  # will trigger bulldozer.
+  merge_method:
+    # "method" defines the merge method. The available options are "merge",
+    # "rebase", "squash", and "ff-only".
+    - method: squash
+      trigger:
+        # All methods from merge/trigger are supported. Additionally, the
+        # following additional methods are provided:
+
+        # Pull requests which a number of commits less than or equal to this value are added to the trigger.
+        max_commits: 3
 
   # "options" defines additional options for the individual merge methods.
   options:
@@ -152,7 +172,7 @@ merge:
       # If "body" is "pull_request_body", then the commit message will be the
       # part of the pull request body surrounded by "message_delimiter"
       # strings. This is disabled (empty string) by default.
-      message_delimiter: ==COMMIT_MSG==
+    message_delimiter:bulldozer/config_v1.go ==COMMIT_MSG==
 
   # "required_statuses" is a list of additional status contexts that must pass
   # before bulldozer can merge a pull request. This is useful if you want to
--- a/README.md
+++ b/README.md
@@ -140,9 +140,10 @@ merge:
   # Allows the merge method that is used when auto-merging a PR to be different
   # based on trigger criteria. The first method where ALL triggers match will
   # be used. Otherwise, the method specified previously in merge/method will be # used.
-  #
-  # Note: ALL trigger criteria must match, unlike merge/trigger where ANY match 
+  # - ALL trigger criteria must match, unlike merge/trigger where ANY match 
   # will trigger bulldozer.
+  # - This will override any branch_method logic if one of the methods is
+  # triggered
   merge_method:
     # "method" defines the merge method. The available options are "merge",
     # "rebase", "squash", and "ff-only".
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -44,9 +44,9 @@ type MergeConfig struct {
 	DeleteAfterMerge       bool `yaml:"delete_after_merge"`
 	AllowMergeWithNoChecks bool `yaml:"allow_merge_with_no_checks"`
 
-	Method  MergeMethod        `yaml:"method"`
-	Methods []ConditionalMerge `yaml:"merge_method"`
-	Options MergeOptions       `yaml:"options"`
+	Method       MergeMethod              `yaml:"method"`
+	MergeMethods []ConditionalMergeMethod `yaml:"merge_method"`
+	Options      MergeOptions             `yaml:"options"`
 
 	BranchMethod map[string]MergeMethod `yaml:"branch_method"`
 
@@ -59,7 +59,7 @@ type MergeOptions struct {
 	Squash *SquashOptions `yaml:"squash"`
 }
 
-type ConditionalMerge struct {
+type ConditionalMergeMethod struct {
 	Method  MergeMethod `yaml:"method"`
 	Trigger Signals     `yaml:"trigger"`
 }
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -45,7 +45,7 @@ type MergeConfig struct {
 	AllowMergeWithNoChecks bool `yaml:"allow_merge_with_no_checks"`
 
 	Method  MergeMethod        `yaml:"method"`
-	Methods []ConditionalMerge `yaml:"methods"`
+	Methods []ConditionalMerge `yaml:"merge_method"`
 	Options MergeOptions       `yaml:"options"`
 
 	BranchMethod map[string]MergeMethod `yaml:"branch_method"`
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -44,8 +44,9 @@ type MergeConfig struct {
 	DeleteAfterMerge       bool `yaml:"delete_after_merge"`
 	AllowMergeWithNoChecks bool `yaml:"allow_merge_with_no_checks"`
 
-	Method  MergeMethod  `yaml:"method"`
-	Options MergeOptions `yaml:"options"`
+	Method  MergeMethod        `yaml:"method"`
+	Methods []ConditionalMerge `yaml:"methods"`
+	Options MergeOptions       `yaml:"options"`
 
 	BranchMethod map[string]MergeMethod `yaml:"branch_method"`
 
@@ -56,6 +57,11 @@ type MergeConfig struct {
 
 type MergeOptions struct {
 	Squash *SquashOptions `yaml:"squash"`
+}
+
+type ConditionalMerge struct {
+	Method  MergeMethod `yaml:"method"`
+	Trigger Signals     `yaml:"trigger"`
 }
 
 type SquashOptions struct {
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -45,6 +45,17 @@ func IsPRTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (b
 	return matches, reason, err
 }
 
+// IsMergeMethod returns true if ALL signals are fully matched,
+// false otherwise. Additionally, a description of the reason will be returned.
+func IsMergeMethodTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
+	matches, reason, err := config.MatchesAll(ctx, pullCtx, "triggered")
+	if err != nil {
+		// trigger must always fail closed (no match on error)
+		return false, reason, err
+	}
+	return matches, reason, err
+}
+
 // statusSetDifference returns all statuses in required that are not in actual,
 // accouting for special behavior in GitHub.
 func statusSetDifference(required, actual []string) []string {
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -26,7 +26,7 @@ import (
 // IsPRIgnored returns true if the PR is identified as ignored,
 // false otherwise. Additionally, a description of the reason will be returned.
 func IsPRIgnored(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
-	matches, reason, err := config.Matches(ctx, pullCtx, "ignored")
+	matches, reason, err := config.MatchesAny(ctx, pullCtx, "ignored")
 	if err != nil {
 		// ignore must always fail closed (matches on error)
 		matches = true
@@ -37,7 +37,7 @@ func IsPRIgnored(ctx context.Context, pullCtx pull.Context, config Signals) (boo
 // IsPRTriggered returns true if the PR is identified as triggered,
 // false otherwise. Additionally, a description of the reason will be returned.
 func IsPRTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
-	matches, reason, err := config.Matches(ctx, pullCtx, "triggered")
+	matches, reason, err := config.MatchesAny(ctx, pullCtx, "triggered")
 	if err != nil {
 		// trigger must always fail closed (no match on error)
 		return false, reason, err
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -29,7 +29,7 @@ func IsPRIgnored(ctx context.Context, pullCtx pull.Context, config Signals) (boo
 	matches, reason, err := config.MatchesAny(ctx, pullCtx, "ignored")
 	if err != nil {
 		// ignore must always fail closed (matches on error)
-		return false, reason, err
+		return true, reason, err
 	}
 	return matches, reason, err
 }
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -29,7 +29,7 @@ func IsPRIgnored(ctx context.Context, pullCtx pull.Context, config Signals) (boo
 	matches, reason, err := config.MatchesAny(ctx, pullCtx, "ignored")
 	if err != nil {
 		// ignore must always fail closed (matches on error)
-		matches = true
+		return false, reason, err
 	}
 	return matches, reason, err
 }
--- a/bulldozer/evaluate.go
+++ b/bulldozer/evaluate.go
@@ -45,7 +45,7 @@ func IsPRTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (b
 	return matches, reason, err
 }
 
-// IsMergeMethod returns true if ALL signals are fully matched,
+// IsMergeMethodTriggered returns true if ALL signals are fully matched,
 // false otherwise. Additionally, a description of the reason will be returned.
 func IsMergeMethodTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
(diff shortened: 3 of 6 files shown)


bulldozer/signals.go
```diff
@@ -21,109 +21,322 @@ import (
...
+	}
+
+	_, matches.BranchPatterns, err = s.branchPatternMatches(ctx, pullCtx, tag)
+	if err != nil {
+		return false, "", err
+	}
+
+	logger := zerolog.Ctx(ctx)
+	logger.Debug().Msgf("matches count: %d", matches.size())
+	logger.Debug().Msgf("singals count: %d", s.size())
```
asvoboda:
```suggestion
	logger.Debug().Msgf("signals count: %d", s.size())
```

bulldozer/signals.go
```diff
@@ -21,109 +21,322 @@ import (
...
+// MatchesAll returns true if the pull request matches ALL of the signals. It also
+// returns a description of the match status. The tag argument appears
+// in this description and indicates the behavior (trigger, ignore) this
+// set of signals is associated with.
+func (s *Signals) MatchesAll(ctx context.Context, pullCtx pull.Context, tag string) (bool, string, error) {
+	matches := SignalsMatches{}
+	var err error
+
+	_, matches.Labels, err = s.labelMatches(ctx, pullCtx, tag)
+	// TODO: Collect errors as a list
```
asvoboda: go-muiltierror is a useful tool for this.

bulldozer/signals.go
```diff
@@ -21,109 +21,322 @@ import (
...
+
+// labelMatches determines which label signals match the given PR. It returns:
+// - A list of descriptions for each matched signal
+// - A list of the matched signals
+func (s *Signals) labelMatches(ctx context.Context, pullCtx pull.Context, tag string) ([]string, Signal, error) {
 	logger := zerolog.Ctx(ctx)
+	matches := Signal{}
+	descriptions := []string{}
 
 	labels, err := pullCtx.Labels(ctx)
```
asvoboda, line 169: Kind of unrelated, but I think we could also fast exit here if there aren't any signals present. Same goes for the other functions.

bulldozer/signals.go
```diff
@@ -21,109 +21,322 @@ import (
...
 		if strings.Contains(body, signalSubstring) {
-			return true, fmt.Sprintf("pull request body matches a %s substring: %q", tag, signalSubstring), nil
-		}
-		for _, comment := range comments {
-			if strings.Contains(comment, signalSubstring) {
-				return true, fmt.Sprintf("pull request comment matches a %s substring: %q", tag, signalSubstring), nil
+			matches = append(matches, signalSubstring)
+			descriptions = append(descriptions, fmt.Sprintf("pull request body matches a %s substring: %q", tag, signalSubstring))
+		} else {
+			// TODO: Use a set and remove the else
```
asvoboda: Are you thinking about something like:
```
allStrings := make(map[string][]string) {
"body" : []string{body},
"comments": comments,
}

for k, v := range allStrings {
  if strings.Contains(v, signalSubstring) {
  ...
}
```

asvoboda (commented):
Overall I think this approach makes sense. I'd like to see the PR out of Draft for another review, but looks pretty good so far.

bulldozer/signals.go
derekjobst: I think we'd want to do:
```golang
allStrings := make(map[string][]string) { 
			"body" : []string{body}, 
			"comments": comments, 
		} 
		
		for label, strings := range allStrings { 
			for _, s := range strings {
				if s == signalComment {
					matches = append(matches, signalComment)
					descriptions = append(descriptions, fmt.Sprintf("pull request has a %s %s: %q", tag, label, signalComment))
				}
			}
		}
```

Which seems a little harder to understand than the original. Going to just remove the TODO

bulldozer/merge.go
```diff
@@ -156,6 +156,21 @@ func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConf
 
 	base, head := pullCtx.Branches()
 	mergeMethod := mergeConfig.Method
+	for i := 0; i < len(mergeConfig.Methods); i++ {
```
bluekeyes: style: `for _, method := range mergeConfig.Methods`

bulldozer/merge.go
```diff
@@ -156,6 +156,21 @@ func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConf
...
+		triggered, reason, err := IsMergeMethodTriggered(ctx, pullCtx, conditionalMethod.Trigger)
+		if err != nil {
+			logger.Error().Err(err).Msg("Failed to determine if merge method is triggered")
+			return
+		}
+
+		if triggered {
+			mergeMethod = conditionalMethod.Method
+			logger.Debug().Msgf("%s method is triggered because %s matched", mergeMethod, reason)
+			break
```
bluekeyes, line 173: While `break`ing is fine here, I think it might be nice to extract all of the merge-method selection logic (i.e. this loop, the `BranchMethod` statement, and the validation) into a new helper function. That would make `MergePR` easier to follow and let you return from this loop.

**Edit: I think you addressed this part in an update.** Related, we should also be careful about how this interacts with other options. If one option is selected by the conditional methods and it conflicts with the method selected by the existing `branch_method` logic, which should win? My guess is we want the new conditional merge methods to win.

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+	PRBodySubstrings  Signal `yaml:"pr_body_substrings"`
+	Branches          Signal `yaml:"branches"`
+	BranchPatterns    Signal `yaml:"branch_patterns"`
+	MaxCommits        int `yaml:"max_commits"`
 }
 
-func (s *Signals) Enabled() bool {
+type SignalsMatches struct {
+	Signals
+}
```
bluekeyes: I'm not sure I see the need for this type. It seems to have no extra methods and is not used as a parameter or return value, so the current use could probably use `Signals` directly (but as noted below, I'm not sure that's needed either.)

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+	}
+	return size
 }
 
-// Matches returns true if the pull request meets one or more signals. It also
+func (s *Signals) Enabled() bool {
+	return s.size() > 0
+}
+
+// MatchesAll returns true if the pull request matches ALL of the signals. It also
```
bluekeyes: I initially thought `MatchesAll` meant that the pull request had to match the label condition and the comment condition and etc. But it looks like this also changes the meaning of the individual conditions so that "matching the label condition" now means that all labels in the list must exist.

I'm not sure that behavior is desirable or intuitive, but I could be convinced otherwise. How were you thinking about the match any vs match all behavior?

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+// returns a description of the match status. The tag argument appears
+// in this description and indicates the behavior (trigger, ignore) this
+// set of signals is associated with.
+func (s *Signals) MatchesAll(ctx context.Context, pullCtx pull.Context, tag string) (bool, string, error) {
+	matches := SignalsMatches{}
```
bluekeyes: It looks like this is an accumulator of matches so you can compare against the input at the end. Does it need to be a full struct or would it also work to use an integer counting all the matches?

README.md
```diff
@@ -152,7 +172,7 @@ merge:
       # If "body" is "pull_request_body", then the commit message will be the
       # part of the pull request body surrounded by "message_delimiter"
       # strings. This is disabled (empty string) by default.
-      message_delimiter: ==COMMIT_MSG==
+    message_delimiter:bulldozer/config_v1.go ==COMMIT_MSG==
```
bluekeyes: Unintentional addition?

bulldozer/config_v1.go
```diff
@@ -44,8 +44,9 @@ type MergeConfig struct {
 	DeleteAfterMerge       bool `yaml:"delete_after_merge"`
 	AllowMergeWithNoChecks bool `yaml:"allow_merge_with_no_checks"`
 
-	Method  MergeMethod  `yaml:"method"`
-	Options MergeOptions `yaml:"options"`
+	Method  MergeMethod        `yaml:"method"`
+	Methods []ConditionalMerge `yaml:"merge_method"`
```
bluekeyes: Calling this `merge_method` in the configuration but `Methods` in the code seems a bit confusing.

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+	}
+
+	_, matches.MaxCommits, err = s.maxCommitsMatches(ctx, pullCtx, tag)
+	if err != nil {
+		return false, "", err
+	}
+
+	logger := zerolog.Ctx(ctx)
+	logger.Debug().Msgf("matches count: %d", matches.size())
+	logger.Debug().Msgf("signals count: %d", s.size())
```
bluekeyes: These should probably be a single message so they stick together.

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+	}
+
+	return false, fmt.Sprintf("pull request does not match the %s", tag), nil
+}
+
+// labelMatches determines which label signals match the given PR. It returns:
+// - A list of descriptions for each matched signal
+//   - These will only include the first item the signal matched when there is more than one
+// - A list of the matched signals
+func (s *Signals) labelMatches(ctx context.Context, pullCtx pull.Context, tag string) ([]string, Signal, error) {
```
bluekeyes: Following on from my comment above about all/any, this signature feels a bit weird:

* Multiple descriptions are returned but they're either all discarded or only the first is used
* Multiple matching signals are returned, but ultimately callers only care about the count

Going into this, I was expecting these to have the `(bool, string, error)` signature of the callers, which admittedly only works if the "all" behavior applies when combining conditions, not within a single condition.

If we do keep the current behavior, `(int, string, error)` might be a better signature.

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+
 	for _, signalLabel := range s.Labels {
 		for _, label := range labels {
 			if strings.EqualFold(signalLabel, label) {
-				return true, fmt.Sprintf("pull request has a %s label: %q", tag, signalLabel), nil
+				matches = append(matches, signalLabel)
+				descriptions = append(descriptions, fmt.Sprintf("pull request has a %s label: %q", tag, signalLabel))
 			}
 		}
 	}
```
bluekeyes: The new `Signal` type is mostly used as way to distinguish string slices when a function returns more than one, but I think it's a missed opportunity to abstract some of this functionality. `Signal` could define a function (or functions) to determine if an input matches. It would probably make sense to define multiple types of `Signal` for the different matches (literal, case-insensitive, regexp), but multiple methods on a single type could work too.

bulldozer/signals.go
```diff
@@ -21,109 +21,389 @@ import (
...
+
+	if len(commits) < s.MaxCommits {
+		matches = s.MaxCommits
+		descriptions = append(descriptions, fmt.Sprintf("pull request has %q commits, which is less than the maximum of %q", len(commits), s.MaxCommits))
+	}
+
+	if len(commits) == s.MaxCommits {
+		matches = s.MaxCommits
+		descriptions = append(descriptions, fmt.Sprintf("pull request has %q commits, which is the maximum", len(commits)))
+	}
```
bluekeyes: Does this need to be a separate condition? It seems like we could write a single description that gives the necessary information in both cases.

bulldozer/signals_test.go
```diff
@@ -137,7 +137,73 @@ func TestSignalsMatches(t *testing.T) {
...
+				assert.True(t, matches, "expected pull request to match, but it didn't")
+			} else {
+				assert.False(t, matches, "expected pull request to not match, but it did")
+			}
+			assert.Equal(t, test.Reason, reason)
+		})
+	}
+}
+
+func TestSignalsMatchesAll(t *testing.T) {
```
bluekeyes: Thanks for adding a test to cover the max commits behavior! I think this also needs to cover the "matches all" behavior, which is another important change.

bluekeyes (commented):
This is looking good! I highlighted a few areas for cleanup/refactors and I'm also a bit confused by the "matches all" behavior, so it would be good to discuss that.

bulldozer/merge.go
derekjobst: The `merge_method` logic should override `branch_method` if there is a match. I made an additional note about this in the PR.

Good point about breaking out the logic. I pulled it out earlier to try and test the method, so I'll bring part of those changes in.

bulldozer/signals.go
derekjobst: Good point, removed

bulldozer/signals.go
derekjobst: I added this new matching logic to accommodate the new `max_commits` condition. Specifying a max number of commits in a PR was requested as a condition which would be used in conjunction with other trigger criteria. Since the merge methods are defined in a list which can be used to create "OR" logic, it felt logical to just "AND" all the triggers.

You are right this also changes the way triggering each signal works. I didn't consider that as a major issue when I was developing this, but I agree it's confusing. I'll go ahead and revert that.

bulldozer/signals.go
derekjobst: it doesn't. I added it there in case it was needed for logging or errors later on. It might be helpful if we want to add more details about matches to the UI but that might be premature.

bulldozer/config_v1.go
derekjobst: Changed to `MergeMethods`. I used `merge_method` because of `branch_method`, although `merge_methods` makes more sense. 

Since `branch_method` is going away perhaps `merge_methods`  would be best.

bulldozer/signals.go
derekjobst: Removed, used for debugging

README.md
```diff
@@ -126,13 +126,34 @@ merge:
...
   # will be used for PRs targeting that branch. The valid values are the same as for the "method" key.
   # Note: If the target branch does not match any of the specified keys, the "method" key is used instead.
   branch_method:
     develop: squash
     master: merge
+  ##### branch_method has been DEPRECATED in favor of merge_method #####
+
+  # Allows the merge method that is used when auto-merging a PR to be different
+  # based on trigger criteria. The first method where ALL triggers match will
+  # be used. Otherwise, the method specified previously in merge/method will be # used.
```
bluekeyes: I usually use quotes and dots to refer to specific keys: `"merge.method"` or just `"method"` as in the `branch_method` docs. Also looks like the last word should be on a new line (or delete the `#`)

README.md
```diff
@@ -152,7 +173,7 @@ merge:
       # If "body" is "pull_request_body", then the commit message will be the
       # part of the pull request body surrounded by "message_delimiter"
       # strings. This is disabled (empty string) by default.
-      message_delimiter: ==COMMIT_MSG==
+    message_delimiter: ==COMMIT_MSG==
```
bluekeyes: Was this indentation actually incorrect?

bulldozer/evaluate.go
```diff
@@ -37,7 +37,18 @@ func IsPRIgnored(ctx context.Context, pullCtx pull.Context, config Signals) (boo
 // IsPRTriggered returns true if the PR is identified as triggered,
 // false otherwise. Additionally, a description of the reason will be returned.
 func IsPRTriggered(ctx context.Context, pullCtx pull.Context, config Signals) (bool, string, error) {
-	matches, reason, err := config.Matches(ctx, pullCtx, "triggered")
+	matches, reason, err := config.MatchesAny(ctx, pullCtx, "triggered")
+	if err != nil {
+		// trigger must always fail closed (no match on error)
+		return false, reason, err
```
bluekeyes, line 43: Not really part of your change, but I just noticed that `IsPRIgnored` doesn't match the implementation of these other two `Is__` methods - might be nice to have them do the same thing (either update `matches` or return directly)

bulldozer/merge.go
```diff
@@ -149,21 +149,48 @@ func (m *PushRestrictionMerger) DeleteHead(ctx context.Context, pullCtx pull.Con
...
 
 	if branchMergeMethod, ok := mergeConfig.BranchMethod[base]; ok {
 		mergeMethod = branchMergeMethod
 	}
+
+	for _, method := range mergeConfig.MergeMethods {
+		triggered, reason, err := IsMergeMethodTriggered(ctx, pullCtx, method.Trigger)
+		if err != nil {
+			logger.Error().Err(err).Msg("Failed to determine if merge method is triggered")
+			return "", err
```
bluekeyes: Since this returns an error, we should probably log it in the caller, which does not return an error, instead of here.

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
 	"strings"
 
 	"github.com/palantir/bulldozer/pull"
+	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
 )
 
+type Signal interface {
+	Enabled() bool
+	// Determine if the signal has values assigned to it and should be considered when matching
```
bluekeyes, line 29: style: these should appear above the interface methods, like normal documentation comments

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
...
-	PRBodySubstrings  []string `yaml:"pr_body_substrings"`
-	Branches          []string `yaml:"branches"`
-	BranchPatterns    []string `yaml:"branch_patterns"`
+	Labels            `yaml:"labels"`
+	CommentSubstrings `yaml:"comment_substrings"`
+	Comments          `yaml:"comments"`
+	PRBodySubstrings  `yaml:"pr_body_substrings"`
+	Branches          `yaml:"branches"`
+	BranchPatterns    `yaml:"branch_patterns"`
+	MaxCommits        `yaml:"max_commits"`
```
bluekeyes: While a bit more verbose, I'd make these named fields instead of embedded structs for clarity, since that is how they are used, e.g.:

```go
Labels Labels `yaml:"labels"`
```

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
...
+}
+
+func (signal *Labels) Enabled() bool {
```
bluekeyes: Slices function similarly to pointers in that they are a small amount of data that refers to more data elsewhere in memory. There's usually no reason to take a pointer to a slice unless you need to modify the slice.

For that reason, I think you'll want to define all of these methods (along with `Matches`) on the value type rather than on the pointer type. This will also save you a bunch of dereferencing.

Same comments apply to the `MaxCommit` functions too.

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
...
+		&s.MaxCommits,
+	}
+
+	for _, signal := range signals {
+		matches, _, err := signal.Matches(ctx, pullCtx, tag)
+		if err != nil {
+			return false, "", err
+		}
+
+		if signal.Enabled() && !matches {
```
bluekeyes: I see you added debug logging for when `Matches` is called but the signal is disabled. I'm worried that could get quite verbose for conditions that only use one or two signals. It might be better to make this `Enabled` check before calling `Matches` and then skip the signal if it is not enabled.

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
...
+}
+
+// Matches Determines which label signals match the given PR. It returns:
+// - A boolean to indicate if a signal matched
+// - A description of the first matched signal
+func (signal *Labels) Matches(ctx context.Context, pullCtx pull.Context, tag string) (bool, string, error) {
 	logger := zerolog.Ctx(ctx)
 
+	if !signal.Enabled() {
+		logger.Debug().Msgf("No label signals have been provided to match against")
```
bluekeyes: Even if you keep the structure of `MatchesAll` the same, I think we may want to remove these log messages to reduce noise. I'm not sure how useful these will be to someone debugging a problem, especially when only a few signal types are in use at any given time.

bulldozer/merge.go
```diff
@@ -149,21 +149,48 @@ func (m *PushRestrictionMerger) DeleteHead(ctx context.Context, pullCtx pull.Con
 	return m.Normal.DeleteHead(ctx, pullCtx)
 }
 
-// MergePR merges a pull request if all conditions are met. It logs any errors
-// that it encounters.
-func MergePR(ctx context.Context, pullCtx pull.Context, merger Merger, mergeConfig MergeConfig) {
+// DetermineMergeMethod determines which merge method to use when merging the PR
+func DetermineMergeMethod(ctx context.Context, pullCtx pull.Context, mergeConfig MergeConfig) (MergeMethod, error) {
```
bluekeyes, line 153: Thanks for extracting this, much easier to read now!

bulldozer/signals.go
```diff
@@ -21,62 +21,186 @@ import (
...
+		&s.Labels,
+		&s.CommentSubstrings,
+		&s.Comments,
+		&s.PRBodySubstrings,
+		&s.Branches,
+		&s.BranchPatterns,
+		&s.MaxCommits,
+	}
+
+	for _, signal := range signals {
```
bluekeyes, line 111: I like the `Signal` interface - this loop makes the logic of these matching methods easy to follow.

bulldozer/signals_test.go
```diff
@@ -137,7 +137,140 @@ func TestSignalsMatches(t *testing.T) {
...
```
bluekeyes, line 257: Thanks for adding a test for this!

bluekeyes (commented):
The structure of this looks great now. I had a few more style comments, but then we should be good to merge.

bulldozer/signals.go
derekjobst: Ok, sounds good

bulldozer/signals.go
derekjobst: Ok, will do. Perhaps it would be helpful to add a `Signal` suffix to each type to differentiate the name and the type?

README.md
derekjobst: That must have been a mistake, good catch

bulldozer/signals.go
bluekeyes: Yeah, if you want to, that sounds fine - I don't have a strong opinion either way.

bulldozer/signals_test.go
derekjobst: For sure!

bulldozer/signals.go
```diff
@@ -105,13 +105,15 @@ func (s *Signals) MatchesAll(ctx context.Context, pullCtx pull.Context, tag stri
 	}
 
 	for _, signal := range signals {
-		matches, _, err := signal.Matches(ctx, pullCtx, tag)
-		if err != nil {
-			return false, "", err
-		}
+		if signal.Enabled() {
```
bluekeyes, line 112: Just checking: if no triggers/signals are specified in the configuration (i.e. `Enabled()` always returns false), `MatchesAll` will return true. Is that the intended behavior? I can see valid arguments for both `true` and `false` here.

bluekeyes (approved):
Thanks for all the fixes. One more question, then let's try it out!

bulldozer/signals.go
asvoboda: I think documenting the behaviour might be nice either way. Similar to #217, we may want to either provide an option to opt out of the behaviour. I think the current behaviour is good for now though.

derekjobst: If there are no triggers/signals, then yes MatchesAll will be true. I agree this could go either way, although if there are no signals/triggers defined that is an invalid config as it is currently defined. I'll go ahead and make that return `false` and update the docs.

Issue #276, opened 2022-01-19
Sporadic 403 "You have exceeded a secondary rate limit” errors


daiglebg:
Our org currently has Bulldozer configured for 30 repos and 300+ open pull requests. We are seeing sporadic “403 You have exceeded a secondary rate limit” errors in our Bulldozer logs. Is it possible that Bulldozer is running too many concurrent api requests and github is throwing the 403 error? An example of the complete error is:

```
{
  "level": "error",
  "rid": "c7it9upsbsbhkcad330g",
  "github_event_type": "push",
  "github_delivery_id": "26036a76-77d6-11ec-9293-e49c7563bcc0",
  "github_installation_id": 6133582,
  "github_repository_owner": "OUR_ORG",
  "github_repository_name": "our_repo",
  "github_pr_num": 2424,
[...]

PR #278, opened 2022-01-22
Add delay between pull request updates

bluekeyes:
In projects with many open PRs using the automatic updates feature,
Bulldozer can trigger GitHub's secondary rate limits by updating (and
possibly querying) pull requests too quickly. To avoid this, add some
delay between operations.

For queries, use a fixed 250ms delay for now. This shouldn't slow things
down too much, but does prevent Bulldozer from blasting through PRs as
fast as possible.

For updates, use an exponential delay that starts at 1s and maxes out at
60s. I believe the secondary limits (at least for GitHub Enterprise) are
computed over a minute, so this will limit us to one update per minute
for large batches. I don't think this will be prohibitively slow, but we
could explore making this configurable in the future.

As part of this, refactor the logic so that we don't block queries on
previous updates.

This should help with the issues reported in #276.
Requires #277 to merge first.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/c2h5oh/datasize"
 	"github.com/palantir/bulldozer/bulldozer"
+	"github.com/palantir/bulldozer/server/handler"
 	"github.com/palantir/go-baseapp/baseapp"
 	"github.com/palantir/go-baseapp/baseapp/datadog"
 	"github.com/palantir/go-githubapp/githubapp"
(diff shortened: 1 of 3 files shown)


asvoboda (approved):
While I think the approach of sleeping ahead of time works, it'd be nice to consider handling github rate limit errors and retrying with delays. I know the latter approach needs to handle state a bit differently from what we're doing today, so maybe its not a great use of time atm.

bluekeyes:
Based on @daiglebg's testing in #276, I've changed this to use a fixed 2 second delay for updates. With the maximum value that low, exponential backoff hit the limit so fast that the complexity didn't seem worthwhile. To compensate, I've added configuration options for both delays so that users can increase them or disable them if needed.

The configuration is undocumented for now, as I think most people should use the defaults.

asvoboda:
@bluekeyes are you still interested in merging this?

PR #282, opened 2022-02-10
Add logging for additional webhook events

derekjobst:
Currently only the `push` and `pull_request` events have debug logging when a webhook is received. This is helpful for determining the trigger for bulldozer evaluation when those events occur. This PR includes similar logging for `status`, `issue_comment`, and `pull_reuqest_review` events.

### DIFF
--- a/server/handler/pull_request_review.go
+++ b/server/handler/pull_request_review.go
@@ -45,6 +45,8 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
 	installationID := githubapp.GetInstallationIDFromEvent(&event)
 	ctx, logger := githubapp.PreparePRContext(ctx, installationID, repo, number)
 
+	logger.Debug().Msgf("received pull_request_review %s event", event.GetAction())
+
 	client, err := h.ClientCreator.NewInstallationClient(installationID)
 	if err != nil {
 		return errors.Wrap(err, "failed to instantiate github client")
--- a/server/handler/pull_request_review.go
+++ b/server/handler/pull_request_review.go
@@ -45,7 +45,7 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
 	installationID := githubapp.GetInstallationIDFromEvent(&event)
 	ctx, logger := githubapp.PreparePRContext(ctx, installationID, repo, number)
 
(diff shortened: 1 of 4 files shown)


server/handler/pull_request_review.go
```diff
@@ -45,6 +45,8 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
 	installationID := githubapp.GetInstallationIDFromEvent(&event)
 	ctx, logger := githubapp.PreparePRContext(ctx, installationID, repo, number)
 
+	logger.Debug().Msgf("received pull_request_review %s event", event.GetAction())
```
derekjobst: The `action` field is always going to be submitted in this case, so it might not be all that useful. However, keeping the value ensures the messages match those for other event types.

Would it be helpful to include the user information here?

server/handler/status.go
```diff
@@ -39,12 +39,17 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+	checkName := event.GetName()
+	checkState := event.GetState()
+
 	owner := repo.GetOwner().GetLogin()
 	repoName := repo.GetName()
 	installationID := githubapp.GetInstallationIDFromEvent(&event)
 	ctx, logger := githubapp.PrepareRepoContext(ctx, installationID, repo)
 
-	if event.GetState() != "success" {
+	logger.Debug().Msgf("Received status event for %s with state %s", checkName, checkState)
```
bluekeyes: Since you're adding/changing these events, could we make sure they all use a consistent message, i.e. same capitalization, roughly the same text, etc?

bluekeyes (commented):
I think this is fine if it helps with debugging, but note that `go-githubapp` already includes some [basic logging for all webhook deliveries](https://github.com/palantir/go-githubapp/blob/develop/githubapp/dispatcher.go#L188). I guess this isn't sufficient because it does not include the action field from the payload and is logged regardless of whether Bulldozer is going to act on the event?

That feature from `go-githubapp` also means that most log messages should already be tagged with the originating event type, although again, it will be missing the `action` field.

server/handler/status.go
derekjobst: Capitalization fixed. Changed this message to include the state where other messages include the action. e.g. "Received status success event from policy-bot: develop"

bluekeyes (approved):

PR #285, opened 2022-02-16
Update dist step to use go 1.17

asvoboda:


bluekeyes:
This looks fine, but I realized we can update our workflows to use the `.palantir/go-version` file, which feels like a better solution: https://github.com/palantir/policy-bot/pull/386

bluekeyes:
Closing in favor of #286.

PR #294, opened 2022-03-10
Skip updating pull requests with a server override

asvoboda:


server/config.go
```diff
@@ -66,6 +66,8 @@ type Options struct {
 	DefaultRepositoryConfig *bulldozer.Config `yaml:"default_repository_config"`
 
 	ConfigurationV0Paths []string `yaml:"configuration_v0_paths"`
+
+	DisableUpdateFeature bool `yaml:"disable_update_feature"`
```
asvoboda, line 70: I'll add a readme entry for this

README.md
```diff
@@ -416,6 +416,20 @@ The upgrade process is therefore to deploy the latest version of bulldozer with
...
+can cause lots of pressure on CI build systems and Github when there are many pull requests
+open on a single repository.
+
+To reduce pressure on CI systems and Github, the update feature can be disabled at the
+server level by specifying the following server option:
+
+```yaml
+options:
+  disable_update_feature: true
+
```
bluekeyes: Looks like this is missing the closing code block marker.

server/handler/pull_request.go
```diff
@@ -69,14 +69,18 @@ func (h *PullRequest) Handle(ctx context.Context, eventType, deliveryID string,
...
 	if event.GetAction() == "labeled" || event.GetAction() == "opened" {
-		base, _ := pullCtx.Branches()
-		didUpdatePR, err := h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, config, pr, base)
-		if err != nil {
-			logger.Error().Err(errors.WithStack(err)).Msg("Error updating pull request")
-		}
-
-		if didUpdatePR {
-			return nil
+		if !h.DisableUpdateFeature {
```
bluekeyes: Minor style thing: I might invert this condition so that the skip/log happens in the `if` and the update happens in the `else`.

server/config.go
bluekeyes, line 70: Thinking out loud here: are there any other features we'd want this for in the future? Would it make sense to create some kind of feature flag configuration section or are individual options fine for the one or two things this might apply to?

asvoboda: I think a feature flag section might be interesting. I think the `options` block kind of fits that purpose already?

bluekeyes (approved):

PR #295, opened 2022-03-10
Add `auto_merge` trigger

### LINKED ISSUE #289: New trigger idea: auto_merge
Hello!

I recently noticed that GitHub started exposing pull request's `auto_merge` status. What do you think of adding support for that as a new trigger?

It would actually make a lot of sense in our use case – the main thing that we need from `bulldozer` is auto-updating pull requests when they've been marked as 'ready to merge'. However using native GitHub functionality to mark PRs as ready seems more natural than labels.

Happy to work on a PR!

heuels:
This PR adds a new trigger type – `auto_merge`. GitHub REST API v3 exposes this status on the PR object, and it can be a pretty good indicator of merge readiness, a good alternative to `ready to merge` label for simple cases.

Please let me know if anything's amiss in my PR – this is me trying Go for the first time 😅 

Closes #289.

cc @bluekeyes

### DIFF
--- a/bulldozer/signals.go
+++ b/bulldozer/signals.go
@@ -40,6 +40,7 @@ type PRBodySubstringsSignal []string
 type BranchesSignal []string
 type BranchPatternsSignal []string
 type MaxCommitsSignal int
+type AutoMergeSignal bool
 
 type Signals struct {
 	Labels            LabelsSignal            `yaml:"labels"`
@@ -49,6 +50,7 @@ type Signals struct {
 	Branches          BranchesSignal          `yaml:"branches"`
 	BranchPatterns    BranchPatternsSignal    `yaml:"branch_patterns"`
 	MaxCommits        MaxCommitsSignal        `yaml:"max_commits"`
+	AutoMerge         AutoMergeSignal         `yaml:"auto_merge"`
 }
 
 func (signal LabelsSignal) Enabled() bool {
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @heuels! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bulldozer/signals.go
```diff
@@ -140,6 +148,7 @@ func (s Signals) MatchesAny(ctx context.Context, pullCtx pull.Context, tag strin
 		&s.PRBodySubstrings,
 		&s.Branches,
 		&s.BranchPatterns,
```
heuels, line 150: I noticed that `MaxCommits` signal is missing here. Is this intentional?

bluekeyes: Good catch, but I believe this is intentional for now. When we added the `MaxCommits` signal, it was specifically for the `mege_method` block, which is the only block that uses `MatchesAll` right now. We may want to change that in the future, but I think we can leave it alone for now.

bluekeyes (approved):
Thanks for implementing this. I wouldn't have guessed it was your first time using Go!

If you'd like to test it out, you can use the `palantirtechnologies/bulldozer:snapshot` Docker image that will publish once this merges. Otherwise, I think we might tag a release in the next few days.

PR #307, opened 2022-05-02
Support required statuses for Bulldozer to update PRs

### LINKED ISSUE #259: Option to only update when certain required status are passing
Similar to how `merge` can be configured to not take action unless all `required_statuses` are passing or neutral, it would be helpful if `update` could be configured to not take action unless all `required_statuses` are passing or neutral. This would help prevent failing builds from being continuously updated and wasting CI resources.

Considerations for implementation:
1. `required_statuses` should be separately configured for `update` and `merge`
1. `required_statuses` for `update` should be optional
1. The `update` support for `required_statuses` should reuse the configuration interfaces and code implementations for the equivalent support in `merge` wherever possible, to help keep things consistent between the two and reduce maintenance over time.

Open questions:
1. Should `trigger` conditions override the `required_statuses` like they do for `ignore_drafts` (added in https://github.com/palantir/bulldozer/pull/256)? This would allow users to manually override failing statuses to keep updates happening (in case an update is expected to resolve a failing status)

wrslatz:
### Summary

Closes https://github.com/palantir/bulldozer/issues/259

1) Add `required_statuses` to Bulldozer update config
    - Refactored `ShouldUpdatePR` to `evaluate.go` to be alongside `ShouldMergePR` and reuse some status checks code
    - No tests exist anymore in `update_test.go` since they all covered `ShouldUpdatePR` and were refactored to `evaluate_test.go`
2) Enable update processing when status or check run events are handled
3) Additional changes
    - Improve logging to differentiate between should merge and should update logs
    - Adjusted config tests to better test provided and default config values
    - Refactor `TestShouldUpdatePR` to table-based tests so it is easier to understand the test cases


wrslatz:
So it looks like status changes *not* being fulfilled is working correctly as implemented 🎉 

However, status changes themselves are not triggering an update, from the local testing I've done. Looks like I need to update the check run `check_run.go`) and status update (`status.go`) handling to consider updates as well as merges like we do for pushes or other pull events.

I'm not sure what implications this might have. Open to any thoughts or feedback.

bluekeyes:
> However, status changes themselves are not triggering an update, from the local testing I've done. Looks like I need to update the check run (`check_run.go`) and status update (`status.go`) handling to consider updates as well as merges like we do for pushes or other pull events.

Yes, this sounds correct. Because statuses are pretty frequent events, it's important to filter them if we're not already doing so. At minimum, we should exit the handler immediately for non-successful statuses. Offhand, I'm not sure if there are any other optimizations to make here.

wrslatz:
> Yes, this sounds correct. Because statuses are pretty frequent events, it's important to filter them if we're not already doing so. At minimum, we should exit the handler immediately for non-successful statuses.

It looks like both already handle this logic
https://github.com/palantir/bulldozer/blob/c5b53400811456a5cdf4eea3c1af9b60f8ebf27a/server/handler/status.go#L52-L55
https://github.com/palantir/bulldozer/blob/c5b53400811456a5cdf4eea3c1af9b60f8ebf27a/server/handler/check_run.go#L47-L50

> Offhand, I'm not sure if there are any other optimizations to make here.

It looks like both of these handlers act on every open PR when processing
https://github.com/palantir/bulldozer/blob/c5b53400811456a5cdf4eea3c1af9b60f8ebf27a/server/handler/status.go#L72-L82
https://github.com/palantir/bulldozer/blob/c5b53400811456a5cdf4eea3c1af9b60f8ebf27a/server/handler/check_run.go#L63-L86

Whereas we'd want the update logic to only act on the PR to which the check or status update corresponds?

wrslatz:
Actually, I think I'm misunderstanding how that looping to try to merge works. Perhaps we should try updates in the same places. I'll make an attempt at that now

bluekeyes:
Yeah, both checks and statuses are attached to commits, which could be the HEAD of multiple open PRs (but in practice are usually only part of one.) The loops iterate over all of the open PRs associated with the updated commit, meaning you'll want to check if updates are necessary inside the loop, either before or after checking if merges are necessary.

wrslatz:
I've finished most of the code changes here. I'm rewriting the `TestShouldUpdatePR` to use a table with named props instead of a matrix so it's easier to understand what exactly is being tested.

Do you think we need the full matrix of all combinations of test scenarios? It seems like a _lot_ of tests, and I'm not sure how valuable all of the permutations are vs having tests covering the major scenarios. I can push what I have so far to help with considering, if needed.

bulldozer/evaluate.go
```diff
@@ -102,34 +102,34 @@ func ShouldMergePR(ctx context.Context, pullCtx pull.Context, mergeConfig MergeC
 	if mergeConfig.Ignore.Enabled() {
 		ignored, reason, err := IsPRIgnored(ctx, pullCtx, mergeConfig.Ignore)
 		if err != nil {
-			return false, errors.Wrap(err, "failed to determine if pull request is ignored")
+			return false, errors.Wrap(err, "failed to determine if pull request is ignored for merge")
```
bluekeyes, line 105: Thanks for clarifying/deduplicating these messages!

bulldozer/evaluate.go
```diff
@@ -152,3 +152,58 @@ func ShouldMergePR(ctx context.Context, pullCtx pull.Context, mergeConfig MergeC
...
+	}
+
+	if updateConfig.IgnoreDrafts != nil && *updateConfig.IgnoreDrafts && pullCtx.IsDraft(ctx) {
+		logger.Debug().Msgf("%s is deemed not updateable because PR is in a draft state", pullCtx.Locator())
+		return false, nil
+	}
+
+	requiredStatuses := updateConfig.RequiredStatuses
+
+	if len(requiredStatuses) > 0 {
```
bluekeyes, line 195: Evaluating required statuses here for updates subtly changes the behavior of the property when compared to merges. For merging, required status are checked in addition to other conditions (i.e. must be triggered _and_ have all required statuses.) For updates, this works like `ignore_drafts`: explicitly triggered updates do not require any other conditions, so the required statuses are only for PRs that would update by default.

I don't have a preference for either behavior, but if `required_statuses` are ignored on explicitly triggered updates, I think we should note that in the documentation, like for `ignore_drafts`.

bulldozer/evaluate_test.go
```diff
@@ -326,3 +327,599 @@ func TestShouldMerge(t *testing.T) {
 		assert.True(t, actualShouldMerge)
 	})
 }
+
+func TestShouldUpdatePR(t *testing.T) {
+	ctx := context.Background()
+	tests := []struct {
+		name            string
```
bluekeyes: Thanks for updating these tests! The new structure is much easier to read and follow.

When I write tests like this, I tend to use a map where the test name is the key. Do you have any thoughts on that style vs. including the name as a property?

bulldozer/evaluate_test.go
```diff
@@ -326,3 +327,599 @@ func TestShouldMerge(t *testing.T) {
...
+				SuccessStatusesValue: []string{"status1", "status2"},
+			},
+			updateConfig:    UpdateConfig{
+				IgnoreDrafts: boolVal(true),
+				RequiredStatuses: []string{"status1", "status2"},
+			},
+			expectingUpdate: true,
+		},
+		// Test error handling
+		// TestShouldUpdatePR TODO: add error scenario coverage
```
bluekeyes: If this is a new TODO, it's probably fine to ignore for now, since I don't think we covered the error cases in the old tests either.

server/handler/check_run.go
```diff
@@ -77,8 +77,23 @@ func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, pay
...
+		var didUpdatePR = false
+		if h.DisableUpdateFeature {
+			logger.Debug().Msgf("Skipping updates to pull request due to server configuration override")
+		} else {
+			base, _ := pullCtx.Branches()
+			var err error
+			didUpdatePR, err = h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, config, pr, base)
+			if err != nil {
+				logger.Error().Err(errors.WithStack(err)).Msg("Error updating pull request")
+			}
```
bluekeyes: I think it will be a little simpler (and match the pull request handler) to invert this condition: instead of checking `!didUpdatePR` outside of this branch, check `didUpdatePR` here and continue.
```suggestion
			didUpdatePR, err := h.UpdatePullRequest(logger.WithContext(ctx), pullCtx, client, config, pr, base)
			if err != nil {
				logger.Error().Err(errors.WithStack(err)).Msg("Error updating pull request")
			}
			if didUpdatePR {
				continue
			}
```

bluekeyes (commented):
Thanks for implementing this and especially for updating the tests - I really appreciate you leaving those better than you found them!

Overall this looks good, just a minor behavior question and some style suggestions.

bulldozer/evaluate_test.go
wrslatz: I have no preference either way :+1:

bulldozer/evaluate_test.go
wrslatz: Yep, I'm fine leaving it out. I mainly added it to consider implementing error tests in a future update.

server/handler/check_run.go
wrslatz: Good suggestion, happy to make this change

bulldozer/evaluate.go
wrslatz: I think the difference in behavior for updates is preferred. It aligns better with draft handling like you mentioned. Also, if someone wants to alleviate the update contention, checking required statuses alone seems like a good option. I'll update the docs so it is clearer.

wrslatz: Addressed in https://github.com/palantir/bulldozer/pull/307/commits/0e1a092ce36e68b0f15c0ae11eacd6e568371e09

bulldozer/evaluate_test.go
wrslatz: Addressed in https://github.com/palantir/bulldozer/pull/307/commits/75c6e43aa80995d6007fbc3c6da40f1ae0b12701

server/handler/check_run.go
wrslatz: Addressed for both check_run and status events in https://github.com/palantir/bulldozer/pull/307/commits/92aad8fdb5aae4a22b594ab6902b883ec03908bb

bulldozer/evaluate_test.go
wrslatz: Refactored to map in https://github.com/palantir/bulldozer/pull/307/commits/96baa965556cf8a5c68cc9509ea4398609f14391. I do kind of like how this format looks. Helps compact the cases down a bit with one less line per case.

bluekeyes (approved):

bluekeyes:
Thanks again. I'll try to make a release with this early next week, but feel free to tag me here if that doesn't happen. For now, you can use the `snapshot` tag of the Docker image if you'd like to test it out.

wrslatz:
> Thanks again. I'll try to make a release with this early next week, but feel free to tag me here if that doesn't happen. For now, you can use the `snapshot` tag of the Docker image if you'd like to test it out.

Thank you!

PR #321, opened 2022-06-15
Add draft status to signals

blang:
Adding the draft status to signals allows to ignore pull requests for
auto-merge if they are in draft status.

Example:
```yaml
merge:
    ignore:
        draft: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,19 +118,23 @@ merge:
     # Pull requests with auto merge enabled are added to the trigger.
     auto_merge: true
 
+    # Pull requests in draft status are added to the trigger.
+    draft: true
+
   # "ignore" defines the set of pull request ignored by bulldozer. If the
   # section is missing, bulldozer considers all pull requests. It takes the
   # same keys as the "trigger" section.
   ignore:
     labels: ["do not merge"]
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @blang! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes:
Thanks for adding this! As a signal, this implementation looks good, but I don't think a signal/option is necessary. My understanding is that GitHub does [not allow merging draft pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests), so it should be safe for Bulldozer to always ignore draft PRs when processing merges.

blang:
Currently only tested this with github enterprise 3.3.1 which actually resulted in the PR to be merged despite being in draft state. I was thinking the same but definitely verified this behavior.

PR #323, opened 2022-06-28
Remove 422 error logging

### LINKED ISSUE #253: Avoid logging errors when branches are already deleted
Now that GitHub supports deleting branches on merge, Bulldozer frequently gets a `422` error response when trying to delete a ref. We should suppress these in the logs to avoid confusion, since there's no real problem.

This probably isn't necessary if we end up merging #243.

shravan1k:
Closes #253


derekjobst (approved):
LGTM

Issue #358, opened 2022-11-19
Options can't be configured as ENV var


DepickereSven:
I noticed some properties of the options on bulldozer don't have the capability to be configured as ENV var, on the docker image.
 
The following properties:
```yaml
# Options for application behavior
options:
  # The path within repositories to find the bulldozer.yml config file
  configuration_path: .bulldozer.yml

  # The name of the application. This will affect the User-Agent header
  # when making requests to Github.
  app_name: bulldozer

  # To reduce pressure on CI systems and Github, the update feature can be
  # disabled at the server level by specifying the following server option:
  disable_update_feature: true
```

I would propose that we use the same ENV name as you made possible for [Policy-bot](https://github.com/palantir/policy-bot/blob/develop/config/policy-bot.example.yml), so the options tab would have the prefix `BULLDOZER_OPTIONS_`

So to configure the following vars:
| YAML name   | ENV name   |
|:-:|:-:|
| options.configuration_path  | BULLDOZER_OPTIONS_CONFIGURATION_PATH    |
| options,app_name   | BULLDOZER_OPTIONS_APP_NAME    |
| options,push_restriction_user_token| BULLDOZER_OPTIONS_PUSH_RESTRICTION_USER_TOKEN  |
| options,shared_repository| BULLDOZER_OPTIONS_SHARED_REPOSITORY  |
| options,shared_configuration_path| BULLDOZER_OPTIONS_SHARED_CONFIGURATION_PATH  |
| options,disable_update_feature | BULLDOZER_OPTIONS_DISABLE_UPDATE_FEATURE  |

I would propose this change is a new PR.

DepickereSven:
To not break the logic regarding the prefix: `BULLDOZER_ENV_PREFIX_` I will also support that you can use the prefix instead of `BULLDOZER_OPTIONS_`.

If you configure something with the following prefix => `BULLDOZER_OPTIONS_` and also with `BULLDOZER_ENV_PREFIX_`.
The configured value with `BULLDOZER_ENV_PREFIX_`  will be the actual value then

DepickereSven:
#359 has been merged so this issue can be resolved.

PR #359, opened 2022-11-19
Options can't be configured as ENV var

DepickereSven:
As mentioned in #358 this is the PR for that issue.

So to configure the options of Bulldozer we can use the following ENV vars:
| YAML name   | ENV name   |
|:-:|:-:|
| options.configuration_path  | BULLDOZER_OPTIONS_CONFIGURATION_PATH    |
| options,app_name   | BULLDOZER_OPTIONS_APP_NAME    |
| options,push_restriction_user_token| BULLDOZER_OPTIONS_PUSH_RESTRICTION_USER_TOKEN  |
| options,shared_repository| BULLDOZER_OPTIONS_SHARED_REPOSITORY  |
| options,shared_configuration_path| BULLDOZER_OPTIONS_SHARED_CONFIGURATION_PATH  |
| options,disable_update_feature | BULLDOZER_OPTIONS_DISABLE_UPDATE_FEATURE  |

And the possibility to override with `BULLDOZER_ENV_PREFIX_`.

So that means the following: 
`BULLDOZER_OPTIONS_` and also with `BULLDOZER_ENV_PREFIX_`.
The configured value with `BULLDOZER_ENV_PREFIX_`  will be the actual value then

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -68,9 +68,7 @@ func ParseConfig(bytes []byte) (*Config, error) {
 	}
 	c.Server.SetValuesFromEnv(envPrefix)
 
-	c.Options.SetValuesFromEnv("BULLDOZER_OPTIONS_")
-	// BULLDOZER_ENV_PREFIX_ will override the BULLDOZER_OPTIONS_ values.
-	c.Options.SetValuesFromEnv("BULLDOZER_ENV_PREFIX_")
+	c.Options.SetValuesFromEnv(envPrefix + "OPTIONS_")
 
 	return &c, nil
 }
--- a/server/config.go
+++ b/server/config.go
@@ -15,10 +15,10 @@
 package server
 
 import (
+	"github.com/palantir/bulldozer/server/handler"
 	"os"
 
 	"github.com/c2h5oh/datasize"
-	"github.com/palantir/bulldozer/bulldozer"
 	"github.com/palantir/go-baseapp/baseapp"
 	"github.com/palantir/go-baseapp/baseapp/datadog"
 	"github.com/palantir/go-githubapp/githubapp"
@@ -27,15 +27,13 @@ import (
 )
 
 const (
-	DefaultEnvPrefix               = "BULLDOZER_"
-	DefaultSharedRepository        = ".github"
-	DefaultSharedConfigurationPath = "bulldozer.yml"
+	DefaultEnvPrefix = "BULLDOZER_"
 )
 
 type Config struct {
 	Server  baseapp.HTTPConfig `yaml:"server"`
 	Github  githubapp.Config   `yaml:"github"`
-	Options Options            `yaml:"options"`
+	Options handler.Options    `yaml:"options"`
 	Logging LoggingConfig      `yaml:"logging"`
 	Datadog datadog.Config     `yaml:"datadog"`
 	Cache   CacheConfig        `yaml:"cache"`
@@ -56,20 +54,6 @@ type WorkerConfig struct {
 	QueueSize int `yaml:"queue_size"`
 }
 
-type Options struct {
-	AppName                  string `yaml:"app_name"`
-	PushRestrictionUserToken string `yaml:"push_restriction_user_token"`
-
-	ConfigurationPath       string            `yaml:"configuration_path"`
-	SharedRepository        string            `yaml:"shared_repository"`
-	SharedConfigurationPath string            `yaml:"shared_configuration_path"`
-	DefaultRepositoryConfig *bulldozer.Config `yaml:"default_repository_config"`
-
-	ConfigurationV0Paths []string `yaml:"configuration_v0_paths"`
-
-	DisableUpdateFeature bool `yaml:"disable_update_feature"`
-}
-
 func ParseConfig(bytes []byte) (*Config, error) {
 	var c Config
 	if err := yaml.UnmarshalStrict(bytes, &c); err != nil {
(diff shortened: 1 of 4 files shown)


MstrDC (approved):

server/config.go
```diff
@@ -84,16 +68,9 @@ func ParseConfig(bytes []byte) (*Config, error) {
...
-
-	if c.Options.SharedRepository == "" {
-		c.Options.SharedRepository = DefaultSharedRepository
-	}
-	if c.Options.SharedConfigurationPath == "" {
-		c.Options.SharedConfigurationPath = DefaultSharedConfigurationPath
-	}
+	c.Options.SetValuesFromEnv("BULLDOZER_OPTIONS_")
+	// BULLDOZER_ENV_PREFIX_ will override the BULLDOZER_OPTIONS_ values.
+	c.Options.SetValuesFromEnv("BULLDOZER_ENV_PREFIX_")
```
bluekeyes: `BULLDOZER_ENV_PREFIX` is meant to be a replacement for the `BULLDOZER_` part of the environment variable. After that, the convention I've tried to use for additional app configuration is that the variable name should follow the path in the YAML file. Specifically, I think these lines should be:

```suggestion
	c.Options.SetValuesFromEnv(envPrefix + "OPTIONS_")
```

By default, I would use e.g. `BULLDOZER_OPTIONS_APP_NAME`. If I also set `BULLDOZER_ENV_PREFIX=TEST_`, then I should use `TEST_OPTIONS_APP_NAME`.

server/handler/options.go
```diff
@@ -0,0 +1,85 @@
...
+const (
+	DefaultSharedRepository = ".github"
+	// DefaultConfigurationPath
+	// The default configuration path is the same for the repo as
+	// for the shared config if not configured
+	DefaultConfigurationPath = "bulldozer.yml"
+	DefaultAppName           = "bulldozer"
+)
+
+type Options struct {
```
bluekeyes, line 31: Any particular reason to move this type to the `handler` package or was this to mirror how we did it in `policy-bot`?

server/handler/options.go
```diff
@@ -0,0 +1,85 @@
...
+	"os"
+	"strconv"
+)
+
+const (
+	DefaultSharedRepository = ".github"
+	// DefaultConfigurationPath
+	// The default configuration path is the same for the repo as
+	// for the shared config if not configured
+	DefaultConfigurationPath = "bulldozer.yml"
```
bluekeyes: There's actually a subtle different here. To match the docs, the example configuration, and the (I think) expected behavior, the default configuration path should be the hidden file `.bulldozer.yml`, while the default shared configuration path should be the regular file `bulldozer.yml`. 

This difference allows the shared configuration repository to define its own Bulldozer options that may be different from the configuration used by other repositories in the organization.

bluekeyes (commented):
Thanks for this PR! This was an inconsistency with Bulldozer (especially compared to Policy Bot) that we hadn't gotten a chance to fix yet, so I appreciate the contribution.

I have two small comments on the prefix handling and one of the new defaults, but otherwise this looks good.

server/handler/options.go
DepickereSven: I implemented the way, to mirror how you did it `policy-bot`. Because then it's a bit more separated from the main config. If you want I can move all the main file.

server/config.go
DepickereSven: So that means the following:

| YAML name                           | ENV name                                      | ENV override name                                        |
|:------------------------------------|:----------------------------------------------|:---------------------------------------------------------|
| options.configuration_path          | BULLDOZER_OPTIONS_CONFIGURATION_PATH          | BULLDOZER_ENV_PREFIX_OPTIONS_CONFIGURATION_PATH          |
| options,app_name                    | BULLDOZER_OPTIONS_APP_NAME                    | BULLDOZER_ENV_PREFIX_OPTIONS_APP_NAME                    |
| options,push_restriction_user_token | BULLDOZER_OPTIONS_PUSH_RESTRICTION_USER_TOKEN | BULLDOZER_ENV_PREFIX_OPTIONS_PUSH_RESTRICTION_USER_TOKEN |
| options,shared_repository           | BULLDOZER_OPTIONS_SHARED_REPOSITORY           | BULLDOZER_ENV_PREFIX_OPTIONS_SHARED_REPOSITORY           |
| options,shared_configuration_path   | BULLDOZER_OPTIONS_SHARED_CONFIGURATION_PATH   | BULLDOZER_ENV_PREFIX_OPTIONS_SHARED_CONFIGURATION_PATH   |
| options,disable_update_feature      | BULLDOZER_OPTIONS_DISABLE_UPDATE_FEATURE      | BULLDOZER_ENV_PREFIX_OPTIONS_DISABLE_UPDATE_FEATURE      |

server/handler/options.go
DepickereSven: Yeah thank you for pointing that out, I must have missed that.

bluekeyes (approved):

bluekeyes:
Thanks!

DepickereSven:
@bluekeyes thank you for accepting the change! Any idea when a new release can be created with this feature?

bluekeyes:
I'll try to make a new release of Bulldozer by the end of this week. In the meantime, if you are using our Docker image, you can run the latest `develop` code by using the `snapshot` tag: `docker pull palantirtechnologies/bulldozer:snapshot`

bluekeyes:
This is now available in [1.17.0](https://github.com/palantir/bulldozer/releases/tag/v1.17.0)

PR #370, opened 2023-01-20
Retry merge when github returns 405 with the base branch modified message

### LINKED ISSUE #369: PRs left open when GitHub is too busy
We have been experiencing an intermittent issue where merge requests are rejected by the GitHub API:

```json
{
	"level": "info",
	"rid": "cf3n6dbn37kclfh117a0",
	"github_event_type": "status",
	"github_delivery_id": "3cb85cd2-96e7-11ed-9516-9f680ab3bc27",
	"github_installation_id": 1,
	"github_repository_owner": "acme",
	"github_repository_name": "anvils",
	"github_pr_num": 26725,
	"time": "2023-01-18T04:19:02.713727629Z",
	"message": "Merge rejected due to unsatisfied condition: \"Base branch was modified. Review and try the merge again.\""
}
```

A small percentage of PRs are left unmerged by bulldozer due to the above issue. At this point there are no more triggering events affecting the PR so it never retries. This seems to happen when we have a lot of PRs getting merged in quick succession and is sporadic in nature.

I am happy to contribute a fix for this, but I am not sure what the best approach might be:

1. Retry after a small delay? In the event bulldozer is restarted the PR would once again end up left open, but this would be exceedingly rare and so is acceptable.
2. Periodically scan PRs in a repo? This feels like a can of worms best not opened.

Or if you have any other ideas I would be grateful.

justinwyer:
Resolves #369

### DIFF
--- a/bulldozer/merge_test.go
+++ b/bulldozer/merge_test.go
@@ -15,9 +15,14 @@
 package bulldozer
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"io"
+	"net/http"
 	"testing"
 
+	"github.com/google/go-github/v47/github"
 	"github.com/palantir/bulldozer/pull"
 	"github.com/palantir/bulldozer/pull/pulltest"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


bulldozer/merge_test.go
```diff
@@ -30,9 +35,20 @@ type MockMerger struct {
...
 }
 
 func (m *MockMerger) Merge(ctx context.Context, pullCtx pull.Context, method MergeMethod, msg CommitMessage) (string, error) {
+	if m.GithubStatusCode != 0 {
+		return "", github.CheckResponse(
+			&http.Response{
+				StatusCode: m.GithubStatusCode,
+				Body:       io.NopCloser(bytes.NewReader([]byte(fmt.Sprintf(`{"message": "%s"}`, m.GithubMessage)))),
+			},
+		)
```
bluekeyes: I think we could simplify this by using `github.CheckResponse` to set the the existing `MergeError` field as part of the test, instead of modifying the mock type to understand GitHub statuses.

bulldozer/merge_test.go
```diff
@@ -15,7 +15,12 @@
 package bulldozer
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"github.com/google/go-github/v47/github"
```
bluekeyes: This is the formatting error - it wants this import to be part of non-stdlib block starting on L26.

bluekeyes (commented):
This looks good, and thank you for adding a test! I had one suggestion to simplify the test and then I think this should be ready to merge.

bulldozer/merge_test.go
justinwyer: Thanks, that is a big improvement!

bluekeyes (approved):
Thanks! I'm not sure when the next release will be yet, but you can test this out by using the `palantirtechnologies/bulldozer:snapshot` Docker image, which should publish shortly after this merges.

PR #559, opened 2024-10-30
Filter pull requests pulled by Bulldozer

alankpatel:
Currently we have been getting all pull requests in a given repo but the changes proposed here will filter the results to only retrieve the pull requests we care about via SHA or ref.

### DIFF
--- a/pull/pull_requests.go
+++ b/pull/pull_requests.go
@@ -49,9 +49,7 @@ func ListOpenPullRequestsForRef(ctx context.Context, client *github.Client, owne
 	var results []*github.PullRequest
 	logger := zerolog.Ctx(ctx)
 
-	if strings.HasPrefix(ref, "refs/heads/") {
-		ref = strings.TrimPrefix(ref, "refs/heads/")
-	}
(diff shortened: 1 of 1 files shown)


pull/pull_requests.go
```diff
@@ -26,49 +26,36 @@ import (
...
 func ListOpenPullRequestsForRef(ctx context.Context, client *github.Client, owner, repoName, ref string) ([]*github.PullRequest, error) {
 	var results []*github.PullRequest
 	logger := zerolog.Ctx(ctx)
 
-	openPRs, err := ListOpenPullRequests(ctx, client, owner, repoName)
-
-	if err != nil {
-		return nil, err
+	if strings.HasPrefix(ref, "refs/heads/") {
+		ref = strings.TrimPrefix(ref, "refs/heads/")
```
bluekeyes: nit: `TrimPrefix` returns the unmodified input if it doesn't contain the prefix, so there's usually no need to test with `HasPrefix` before calling `TrimPrefix. If you do care about knowing if the prefix existed or not, you can also use `CutPrefix`.

bluekeyes (approved):

PR #764, opened 2026-06-30
Make MaxPullRequestPollCount configurable via merge config

### LINKED ISSUE #666: Make MaxPullRequestPollCount configurable via YAML
Description: In large organizations with automated PR merges, the current hardcoded limit of 5 retries for the MaxPullRequestPollCount parameter in [bulldozer/merge.go](https://github.com/palantir/bulldozer/blob/develop/bulldozer/merge.go#L31) can become a bottleneck. When bulldozer stops after 5 retries, it may not be sufficient for many workflows, leading to potential merge delays.

Proposal: Please make the MaxPullRequestPollCount parameter configurable via the bulldozer YAML configuration file. This change would allow organizations to customize the retry count based on their specific requirements and workflows.

Benefits:

Flexibility for organizations to set their own retry count.
Reduces likelihood of bulldozer stopping prematurely, improving efficiency.

And I would like work on this issue

sridhar-3009:
## Problem

The retry ceiling for poll-until-mergeable is hardcoded to 5 at compile time (`MaxPullRequestPollCount = 5` in `bulldozer/merge.go`). In organizations running large monorepos or slow CI systems, GitHub's mergeability check often takes longer than 5 × 4 s = 20 s to settle, causing bulldozer to give up and log an error even though the PR would have become mergeable moments later. There is no way to raise this limit without recompiling the server.

## Solution

Add an optional `max_pull_request_poll_count` field to `MergeConfig` (the per-repository `.bulldozer.yml` config). When the field is set to a positive integer, `MergePR` uses that value as the ceiling; when it is unset or zero, the previous default of 5 is used unchanged. This keeps the change fully backward-compatible.

The existing `MaxPullRequestPollCount` constant is retained (renamed to `DefaultMaxPullRequestPollCount`, with an alias for the old name) so any code that references it directly continues to compile.

## Changes

- `bulldozer/config_v1.go`: add `MaxPullRequestPollCount int` field to `MergeConfig`
- `bulldozer/merge.go`: rename constant, read effective limit from `mergeConfig.MaxPullRequestPollCount` at runtime
- `config/examples/standard.bulldozer.v1.yml`: document the new field in a comment

Fixes #666

### DIFF
--- a/changelog/@unreleased/pr-764.v2.yml
+++ b/changelog/@unreleased/pr-764.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: "Make MaxPullRequestPollCount configurable via `max_pull_request_poll_count` in server config"
+  links:
+    - href: https://github.com/palantir/bulldozer/pull/764
+      text: "764"
--- a/bulldozer/config_v1.go
+++ b/bulldozer/config_v1.go
@@ -53,6 +53,11 @@ type MergeConfig struct {
 	// Additional status checks that bulldozer should require
 	// (even if the branch protection settings doesn't require it)
 	RequiredStatuses []string `yaml:"required_statuses"`
+
+	// MaxPullRequestPollCount sets the maximum number of times bulldozer will
+	// poll for a pull request to become mergeable before giving up. If unset
+	// or zero, the default of 5 is used.
+	MaxPullRequestPollCount int `yaml:"max_pull_request_poll_count"`
 }
 
 type MergeOptions struct {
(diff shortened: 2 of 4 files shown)


changelog/@unreleased/pr-764.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: "Make MaxPullRequestPollCount configurable via `max_pull_request_poll_count` in server config"
+  links:
+    - href: https://github.com/palantir/bulldozer/pull/764
+      text: "764"
```
asvoboda, line 6: Please remove this file, its not used anymore and is invalid

Issue #187, opened 2020-06-08
Account for delays in GitHub API updates


bluekeyes:
We've observed a rare race condition where the commit status API does not show the latest values for statuses when processing a status event. This can cause PRs to get stuck unmerged, since there are no more events.

I'm not sure how to fix this, but one option might be to thread through the status payload from the event and merge it with the values from the API. This doesn't help if two different status payloads arrive at the same time and both don't appear in the API response but it might help in the common case.

nmiyake:
FWIW, I believe I observe this intermittently (and more often) in OSS repo basts.

Here are some PRs from an excavator pass where bulldozer failed to merge until it was poked with a manual comment:
* https://github.com/palantir/go-ptimports/pull/53
* https://github.com/palantir/godel-conjure-plugin/pull/78
* https://github.com/palantir/godel-generate-plugin/pull/30
* https://github.com/palantir/godel-goland-plugin/pull/47
* https://github.com/palantir/godel-okgo-asset-unconvert/pull/61
* https://github.com/palantir/outparamcheck/pull/45
* https://github.com/palantir/witchcraft-go-logging/pull/91
* https://github.com/palantir/witchcraft-go-error/pull/42
* https://github.com/palantir/witchcraft-go-params/pull/37

nmiyake:
Just wanted to flag that I don't think this is necessarily rare -- looks like there were at least 5-6 instances of this occurring for a round of excavator PRs, and anecdotally I find that there are often excavator PRs that are unmerged on my repos (wasn't aware of the trick to poke via comment until today)

aeweidne:
I believe we may be hitting this as well, in combination with Policybot.  Policybot UI indicates that the policy matches and is fulfilled, but the commit status seems to be hung and Bulldozer won't merge until someone pokes with a comment.  It appears to happen once we have more than 10-15 PRs being managed at once actively, which seems low

bluekeyes:
@shimmerjs in your case, it sounds like GitHub is still showing a pending status for the Policybot check. Is that correct? If so, leaving a comment also triggers Policybot evaluation again, which will post a new status and trigger Bulldozer. Viewing the UI re-evaluates the policy (without posting a new status), so it's possible for it to disagree with the last posted status.

If you have access to the Policybot logs, there might be something there to explain why the status wasn't updated. 10-15 PRs is very low compared to the number we manage internally, but there are several possible reasons why it might happen: rate limits, saturated processing queue, etc.

[...]

PR #824, opened 2026-09-07
Bind the merge request to the commit that was evaluated

CTWalk:
## Before this PR

`ShouldMergePR` decides whether a pull request may merge using statuses fetched
for the head commit at evaluation time. `defaultMerge` then calls
`PullRequests.Merge` with the pull request **number** and no `sha`, so GitHub
merges whatever the head is when the request arrives.

If a commit is pushed between the check and the merge, the commit that lands is
not the commit that was checked. GitHub's own required status checks cover the
common case — the API refuses the merge, and `attemptMerge` already handles that
`405` — but they do not cover `required_statuses` from `bulldozer.yml`, which
GitHub does not know about. Those are precisely the contexts the option exists
for: *"additional status contexts that must pass before bulldozer can merge…
useful if you want to require extra testing for automated merges."*

`ffOnlyMerge` already binds to the evaluated commit: it passes
`SHA: headCommitSHA` to `UpdateRef` and verifies the ref moved to exactly that
commit. This change gives `defaultMerge` the same guarantee.

## After this PR

`defaultMerge` sets `SHA` on the existing `github.PullRequestOptions` from
`pullCtx.HeadSHA()` — the commit the merge conditions were evaluated against.
GitHub then merges that commit or refuses.

==COMMIT_MSG==
`defaultMerge` called `PullRequests.Merge` without a `sha`, so GitHub merged
whatever the head was when the request arrived rather than the commit whose
statuses were checked. This is invisible for GitHub-required contexts, which the
API enforces at merge time, but `required_statuses` from `bulldozer.yml` are
unknown to GitHub and unenforced by it.

Set `SHA` from `pullCtx.HeadSHA()`, matching what `ffOnlyMerge` already does.
==COMMIT_MSG==

## Possible downsides?

**A pull request whose head moves mid-merge now stops instead of merging.**
GitHub answers `409` on a `sha` mismatch, and `attemptMerge` treats `409` as a
non-retryable stop, so the merge does not happen on that attempt. This is the
intended behaviour change — the alternative is merging an unevaluated commit —
but it is a behaviour change and worth stating plainly.

**It should not leave pull requests stuck.** The push that moved the head produces
its own `pull_request` and status events, so the pull request is evaluated again
against the new commit and merged when it qualifies. Repositories that have seen
stuck-merge symptoms before (#187) may want to watch for this; if maintainers
would prefer the `409` to trigger an immediate re-evaluation rather than waiting
for the next event, I am happy to add that in this PR or a follow-up.

**The log line on that path currently reads** *"Merge rejected due to being
invalid"*, which is accurate but not obviously about a moved head. I left it
alone to keep the diff to the fix; happy to add a distinct message if you want it.

**Scope:** `ffOnlyMerge` is unchanged — it already binds to the evaluated commit
and verifies the result.

If the missing `sha` is deliberate, I'd rather hear the reasoning than have you
carry the PR — happy to close it.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bulldozer/merge.go
+++ b/bulldozer/merge.go
@@ -90,9 +90,16 @@ func (m *GitHubMerger) ffOnlyMerge(ctx context.Context, pullCtx pull.Context) (s
 }
 
 func (m *GitHubMerger) defaultMerge(ctx context.Context, pullCtx pull.Context, method MergeMethod, msg CommitMessage) (string, error) {
+	// Bind the merge to the commit that was evaluated. Without this, GitHub
+	// merges whatever the head is when the request arrives, which may be a
+	// commit pushed after the merge conditions were checked. GitHub rejects a
+	// mismatch with 409, which attemptMerge treats as a non-retryable stop; the
+	// push that moved the head produces its own events, so the pull request is
+	// evaluated again against the new commit.
 	opts := github.PullRequestOptions{
 		CommitTitle: msg.Title,
 		MergeMethod: string(method),
+		SHA:         pullCtx.HeadSHA(),
 	}
 
 	result, _, err := m.client.PullRequests.Merge(ctx, pullCtx.Owner(), pullCtx.Repo(), pullCtx.Number(), msg.Message, &opts)
--- a/bulldozer/merge_test.go
+++ b/bulldozer/merge_test.go
@@ -17,9 +17,11 @@ package bulldozer
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/google/go-github/v90/github"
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/bulldozer](https://github.com/palantir/bulldozer), @CTWalk! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asvoboda:
Thanks for the contribution. Can you also please sign the CLA?

CTWalk:
Flagging a blocker on my side: the CLA signing flow returns a 500 for me.

> Internal Server Error (500)
> An unknown error occured while processing the request.
> ... contact us at opensource@palantir.com

I retried over about 2 days and got the same result each time, so it does not look transient. I emailed opensource@palantir.com today with the details. Happy to rerun the check as soon as the CLA goes through; the branch is rebased and ready otherwise.

asvoboda:
Ack sorry about that, I think something started failing on sept 3. We're looking into it.

asvoboda:
@CTWalk can you give it another shot please?

CTWalk:
@asvoboda Thanks for checking in. I just gave it another shot, but unfortunately it still fails with the same `500 Internal Server Error: An unknown error occurred while processing the request.` 
  Let me know if there are any specific trace details or logs from my attempt that would helpful on your end.

asvoboda:
We unfortunately had the wrong issue identified 2 days ago. We have a new one that was deployed this morning (around 0830 UTC) that should resolve the issue

asvoboda (approved):

PR #831, opened 2026-09-23
Improve bulldozer perf by caching requests and reducing PR calls

asvoboda:
This PR does two things to try and reduce the perf impact of bulldozer on Github (and specifically GHES)

1. empty results are now cached in pull/github_context.go

2. removed the second unnecessary PR fetch in `UpdatePR`


bulldozer/update.go
```diff
@@ -23,15 +23,9 @@ import (
 	"github.com/rs/zerolog"
 )
 
-func UpdatePR(ctx context.Context, pullCtx pull.Context, client *github.Client, updateConfig UpdateConfig, baseRef string) bool {
+func UpdatePR(ctx context.Context, pullCtx pull.Context, client *github.Client, pr *github.PullRequest, updateConfig UpdateConfig, baseRef string) bool {
 	logger := zerolog.Ctx(ctx)
 
-	pr, _, err := client.PullRequests.Get(ctx, pullCtx.Owner(), pullCtx.Repo(), pullCtx.Number())
```
bluekeyes: Are there any branches that may have been relying on this call to get an up-to-date view of the PR compared to what they had before? I'm looking specifically at the check for if the PR is closed on L35.

It's also a bit strange that `pull.Context` itself contains a `*github.PullRequest` object and provides methods to access information about the pull request, but this method takes both. Maybe something to cleanup in a future change.

pull/github_context_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (rc *requestCounter) count(path string) int {
+	rc.mu.Lock()
+	defer rc.mu.Unlock()
+	return rc.counts[path]
+}
+
+func newEmptyResultsContext(t *testing.T) (*GithubContext, *requestCounter) {
+	rc := &requestCounter{counts: make(map[string]int)}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bluekeyes: I'm not sure these tests are that valuable given we don't have any other existing tests for `GitHubContext`. I guess they could help prevent regressions of this particular behavior.

If we do want to test the context, I'd prefer to do it via a mocked `http.Client` like in Policy Bot rather than by starting a test server and making local requests.

bluekeyes (commented):
Caching the empty responses for the duration of the request seems reasonable. I'm not sure how much the update change will help given how little we use that feature, but I guess if we never disabled it in the server, it's still happening on each check run event.

